PR #1, opened 2018-10-04
Improve packaging

enoodle:
* Adding Tox for faster multi-runtime testing
* Adding PBR for automatic versioning
* Auto publishing to pypi

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,2 @@
+Erez Freiberger <erez@epsagon.com>
+ranrib <ran@epsagon.com>
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist
 build
 epsagon.egg-info
 publish.sh
+.tox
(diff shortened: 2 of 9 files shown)


AUTHORS
```diff
@@ -0,0 +1,2 @@
+Erez Freiberger <erez@epsagon.com>
```
ranrib: Use dev@epsagon.com and support@epsagon.com in here

requirements.txt
```diff
@@ -4,3 +4,4 @@ boto3==1.7.35
 simplejson==3.13.2
 six==1.11.0
 future==0.16.0
+Flask
```
ranrib: It's a Dev requirement, not a prod one (whoever not using flask won't need that package)

ranrib:
Looks perfect I love it!

AUTHORS
enoodle: it is autogenerated by PBR by the commits

ranrib: 👍

PR #20, opened 2018-12-01
fix(tox): add support for python 3.7

ranrib:


henprtz:
you need to install python 3.7 in travis

PR #21, opened 2018-12-03
feat(handler.py): support general wrapper

ranrib:
1. Add support for a general wrapper.
2. Improved Linting.
3. Get init params from env

### DIFF
--- a/epsagon/handler.py
+++ b/epsagon/handler.py
@@ -1,0 +1,29 @@
+"""
+Epsagon generic wrapper used only in Lambda environments.
+"""
+
+from .utils import init, import_original_module
+from .wrappers import lambda_wrapper
+
+init()
+ORIGINAL_MODULE = import_original_module()
+
+
+@lambda_wrapper
+def wrapper(event, context):
+    """
+    Generic wrapper for Lambda functions.
+    :param event: Lambda's event
+    :param context: Lambda's context
+    :return: Original handler
+    """
+    original_module, module_path, handler_name = ORIGINAL_MODULE
+
+    try:
+        wrapped_handler = getattr(original_module, handler_name)
+    except AttributeError:
+        raise AttributeError(
+            'No handler {} in module {}'.format(handler_name, module_path)
+        )
+
+    return wrapped_handler(event, context)
--- a/epsagon/utils.py
+++ b/epsagon/utils.py
@@ -97,7 +97,7 @@ def init(
     tracer.initialize(
         token=get_env_or_val('EPSAGON_TOKEN', token),
         app_name=get_env_or_val('EPSAGON_APP_NAME', app_name),
-        collector_url=get_env_or_val('EPSAGON_URL', collector_url),
+        collector_url=get_env_or_val('EPSAGON_COLLECTOR_URL', collector_url),
         metadata_only=(
           (get_env_or_val('EPSAGON_METADATA', '') == 'TRUE') | metadata_only
         ),
--- a/epsagon/utils.py
+++ b/epsagon/utils.py
@@ -3,10 +3,12 @@ Utilities for Epsagon module.
 """
 
 from __future__ import absolute_import
+import os
 from six.moves import urllib
 
 from epsagon.constants import TRACE_COLLECTOR_URL, REGION
 from .trace import tracer
+from .constants import EPSAGON_HANDLER
 
 # Method to URL dict.
 BLACKLIST_URLS = {
@@ -17,6 +19,16 @@ BLACKLIST_URLS = {
         'accounts.google.com',
     ],
 }
+
+
+def get_env_or_val(env_key, value):
+    """
+    return environment variable if exists, otherwise value.
+    :param env_key: environment key
+    :param value: value
+    :return: env or value
+    """
+    return os.getenv(env_key) if os.getenv(env_key) else value
 
 
 def add_data_if_needed(dictionary, name, data):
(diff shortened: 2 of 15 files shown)


epsagon/handler.py
```diff
@@ -0,0 +1,29 @@
...
+
+from .utils import init, import_original_module
+from .wrappers import lambda_wrapper
+
+init()
+ORIGINAL_MODULE = import_original_module()
+
+
+@lambda_wrapper
+def wrapper(event, context):
```
galbash: maybe use the lambda_wrapper only on the user's handler so we won't be in the trace? not sure it's a good idea but think about it.

epsagon/handler.py
```diff
@@ -0,0 +1,29 @@
+"""
+Epsagon generic wrapper used only in Lambda environments.
+"""
+
+from .utils import init, import_original_module
+from .wrappers import lambda_wrapper
+
+init()
```
galbash: think about moving the environment variables reading to here. not sure what makes more sense

epsagon/utils.py
```diff
@@ -19,6 +21,16 @@
 }
 
 
+def get_env_or_val(env_key, value):
```
galbash: you can use `x = os.getenv(env_key) or value`

galbash: so no need for this function at all

epsagon/utils.py
```diff
@@ -80,12 +91,39 @@ def init(token,
...
+        app_name=get_env_or_val('EPSAGON_APP_NAME', app_name),
+        collector_url=get_env_or_val('EPSAGON_COLLECTOR_URL', collector_url),
+        metadata_only=(
+          (get_env_or_val('EPSAGON_METADATA', '') == 'TRUE') | metadata_only
+        ),
+        debug=(get_env_or_val('EPSAGON_DEBUG', '') == 'TRUE') | debug
     )
+
+
+def import_original_module():
```
galbash: look at bootstrap.py

epsagon/utils.py
```diff
@@ -80,12 +91,39 @@ def init(token,
...
+    """
+    Imports original module
+    :return:
```
galbash: docs

epsagon/utils.py
ranrib: Done.

epsagon/utils.py
ranrib: Done.

epsagon/utils.py
ranrib:
```
def _get_handlers(handler, mode):
    lambda_runtime.report_user_init_start()
    init_handler = lambda: None

    """
    This is the old way we were loading modules.
    It was causing intermittent build failures for unknown reasons.
    Using the imp module seems to remove these failures.
    The imp module appears to be more extreme in that it reloads
    the module if it is already loaded, and it likely doesn't use any caches when
    searching for the module but does a full directory search, which is what we want.
    """
    # m = imp.load_module(modname, globals(), locals(), [])

    try:
        (modname, fname) = handler.rsplit('.', 1)
    except ValueError as e:
        fault = wsgi.FaultException("Bad handler '{}'".format(handler), str(e), None)
        request_handler = make_fault_handler(fault)
        lambda_runtime.report_user_init_end()
        return init_handler, request_handler

    file_handle, pathname, desc = None, None, None
    try:
        # Recursively loading handler in nested directories
        for segment in modname.split('.'):
            if pathname is not None:
                pathname = [pathname]
            file_handle, pathname, desc = imp.find_module(segment, pathname)
        if file_handle is None:
            module_type = desc[2]
            if module_type == imp.C_BUILTIN:
                request_handler = make_fault_handler(wsgi.FaultException(
                    "Cannot use built-in module {} as a handler module".format(modname),
                    None,
                    None
                ))
                lambda_runtime.report_user_init_end()
                return init_handler, request_handler
        m = imp.load_module(modname, file_handle, pathname, desc)
    except Exception as e:
        request_handler = load_handler_failed_handler(e, modname)
        lambda_runtime.report_user_init_end()
        return init_handler, request_handler
    finally:
        if file_handle is not None:
            file_handle.close()

    try:
        init_handler = getattr(m, 'init')
    except AttributeError as e:
        pass

    try:
        request_handler = make_final_handler(getattr(m, fname), mode)
    except AttributeError as e:
        fault = wsgi.FaultException("Handler '{}' missing on module '{}'".format(fname, modname), str(e), None)
        request_handler = make_fault_handler(fault)
    lambda_runtime.report_user_init_end()
    return init_handler, request_handler
```

ranrib: Their code is not really nice, and it uses imp that uses importlib, that uses __import__

epsagon/handler.py
ranrib: Done.

epsagon/handler.py
ranrib: I prefer to leave them on the init, because they are relevant even if not using the generic handler.

ranrib:
@galbash done.

epsagon/handler.py
```diff
@@ -0,0 +1,28 @@
...
+    original_module, module_path, handler_name = ORIGINAL_MODULE
+
+    try:
+        wrapped_handler = getattr(original_module, handler_name)
+    except AttributeError:
+        raise AttributeError(
+            'No handler {} in module {}'.format(handler_name, module_path)
+        )
+
+    return lambda_wrapper(wrapped_handler)(event, context)
```
galbash: it's not good that we have to evaluate the wrapper for each execution. let's do it at init

ranrib: Done.

epsagon/handler.py
galbash: this is not what i meant. I meant the code should look like this:
```
wrapper = lambda_wrapper(init_module())
```
and then just set wrapper as the entry point (maybe call it EPSAGON_WRAPPER)

PR #35, opened 2019-01-10
feat(tracer): Limit events storage

henprtz:


epsagon/trace.py
```diff
@@ -185,15 +207,13 @@ def update_runner_with_labels(self):
...
-            (index, ev) for (index, ev) in enumerate(self.events)
-            if ev.origin == 'runner'
+        runner = [
+            event for event in enumerate(self.events())
+            if event.origin == 'runner'
         ][0]
 
         runner.resource['metadata']['labels'] = json.dumps(self.custom_labels)
 
-        self.events[index] = runner
```
enoodle: I understand the reason to delete this line

epsagon/trace.py
```diff
@@ -129,18 +132,28 @@ def load_from_dict(trace_data):
...
     def add_event(self, event):
         """
         Add event to events list.
         :param event: BaseEvent
         :return: None
         """
         event.terminate()
-        self.events.append(event)
+        events = self.events_map.setdefault(event.identifier(), [])
+        if len(events) <= MAX_EVENTS_PER_TYPE:
```
enoodle: maybe a message to tell us that we throw away events?

PR #43, opened 2019-01-26
feat(pynamodb): add support for PynamoDB

ranrib:
🆙

### DIFF
--- a/epsagon/events/pynamodb.py
+++ b/epsagon/events/pynamodb.py
@@ -57,7 +57,7 @@ class PynamoDBEventAdapter(object):
             new_response,
             exception
         )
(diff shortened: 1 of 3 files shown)


epsagon/events/pynamodb.py
```diff
@@ -0,0 +1,63 @@
...
+        event = BotocoreDynamoDBEvent(
+            wrapped,
+            new_instance,
+            new_args,
+            kwargs,
+            start_time,
+            new_response,
+            exception
+        )
+        event.ORIGIN = 'pynamodb'
```
enoodle: travis fails here, because the linter doesn't like the all caps attribute name.

PR #45, opened 2019-02-04
fix(botocore): support update and delete operations in dynamodb

LiadGoren:


epsagon/triggers/aws_lambda.py
```diff
@@ -122,6 +127,22 @@ def __init__(self, start_time, event, context):
             ).hexdigest()
         }
 
+    def _deserialize_item(self, item):
+        """
+
```
galbash: document

LiadGoren: Fixed

PR #46, opened 2019-02-05
feat(timeout): send traces on timeouts

razlo:
Send traces in case of timeout

### DIFF
--- a/epsagon/trace.py
+++ b/epsagon/trace.py
@@ -8,6 +8,7 @@ import time
 import itertools
 import traceback
 import warnings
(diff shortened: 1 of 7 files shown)


epsagon/trace.py
```diff
@@ -137,6 +189,13 @@ def load_from_dict(trace_data):
             trace.add_event(BaseEvent.load_from_dict(event))
         return trace
 
+    def set_runner(self, runner):
+        """
+        TODO
```
ranrib: TODO

PR #57, opened 2019-03-02
feat(httplib2): add support for httplib2

ranrib:
In addition, fixed issue with flask wrapper when DISABLE_EPSAGON=TRUE


epsagon/__init__.py
```diff
@@ -28,7 +28,7 @@ def dummy_wrapper(func):
     step_lambda_wrapper = dummy_wrapper  # pylint: disable=C0103
     azure_wrapper = dummy_wrapper  # pylint: disable=C0103
     python_wrapper = dummy_wrapper  # pylint: disable=C0103
-    EpsagonFlask = dummy_wrapper  # pylint: disable=C0103
```
galbash: is no one using this? you are gonna break someones code (like ours)

galbash: export with both name, maybe add a deprecation warning

epsagon/events/httplib2.py
```diff
@@ -0,0 +1,146 @@
...
+        :param kwargs: wrapt's kwargs
+        :param start_time: Start timestamp (epoch)
+        :param response: response data
+        :param exception: Exception (if happened)
+        """
+
+        super(Httplib2Event, self).__init__(start_time)
+        self.event_id = 'httplib2-{}'.format(str(uuid4()))
+
+        # Params can be set via args or kwargs.
```
galbash: a better approch would be to define an args unroller:
```
def unroller(uri='N/A', method='N/A', body=None, headers=None):
    return uri, method, body, headers

uri, method, body, headers = unroller(*args, **kwargs)
```

epsagon/events/httplib2.py
```diff
@@ -0,0 +1,146 @@
...
+                'request_headers',
+                headers
+            )
+
+        try:
+            if body:
+                add_data_if_needed(
+                    self.resource['metadata'],
+                    'request_body',
+                    simplejson.loads(body)
```
galbash: import simplejson as json. why are we even using this? it just bloats our package

epsagon/events/httplib2.py
```diff
@@ -0,0 +1,146 @@
...
+                )
+        except (TypeError, ValueError):
```
galbash: There is a JSONDecodeError no?

epsagon/events/httplib2.py
```diff
@@ -0,0 +1,146 @@
...
+
+        # Extract only json responses
+        try:
+            if response_body:
+                add_data_if_needed(
+                    self.resource['metadata'],
+                    'response_body',
+                    simplejson.loads(response_body)
+                )
+        except (TypeError, ValueError):
```
galbash: same

epsagon/events/httplib2.py
```diff
@@ -0,0 +1,146 @@
...
+    Factory class, generates Httplib2 event.
+    """
+
+    @staticmethod
+    def create_event(wrapped, instance, args, kwargs, start_time, response,
+                     exception):
+        """
+        Create an event according to the given api_name.
+        """
+        url = kwargs.get('uri', args[0] if len(args) >= 1 else None)
```
galbash: same (unroll)

epsagon/__init__.py
ranrib: The current `EpsagonFlask` breaks the convention actually. Users expects to import `flask_wrapper`

ranrib: EpsagonFlask is not listed in the `__all__`, So I don't think that anyone is doing `from epsagon import EpsagonFlask`

epsagon/events/httplib2.py
ranrib: JSONDecodeError inherits from ValueError

epsagon/events/httplib2.py
ranrib: Done

epsagon/events/httplib2.py
ranrib: Awesome! Done

epsagon/events/httplib2.py
galbash: so catch the specific error... everything inherits from Object maybe just catch it and thats it

epsagon/events/httplib2.py
galbash: same

PR #68, opened 2019-03-06
feat(botocore): support batch get item operation for dynamodb

LiadGoren:


epsagon/events/botocore.py
```diff
@@ -508,6 +509,19 @@ def process_batch_write_op(self):
...
+        Process the batch get item operation.
+        """
+        table_name = list(self.request_data['RequestItems'].keys())[0]
+        self.resource['name'] = table_name
+        keys = [
+            key for key in
+            self.request_data['RequestItems'][table_name]['Keys']
+        ]
+
+        self.resource['metadata']['Keys'] = keys
```
ronyis: Saving all the keys, and especially all the returned items is a bad idea - the trace can easily become too big.
Let's add a maximum of ~10 keys and items.

PR #71, opened 2019-03-15
fix(http_filters): filter amazonaws requests

ranrib:
Since releasing https://github.com/epsagon/epsagon-python/pull/70, we capture all botocore requests twice (once regular, and once as an HTTP req). this PR fixes it.


ranrib:
When you are right, you are right

PR #79, opened 2019-05-07
feat(trace): support more than one tracer to handle concurrent tasks

LiadGoren:


epsagon/modules/tornado.py
```diff
@@ -56,11 +56,11 @@ def after_request(cls, wrapped, instance, args, kwargs):
             ignored = ignore_request(content, '')
             if not ignored and cls.RUNNER:
                 cls.RUNNER.update_response(instance)
-                epsagon.trace.tracer.add_event(cls.RUNNER)
-                epsagon.trace.tracer.send_traces()
-            epsagon.trace.tracer.prepare()
+                trace.add_event(cls.RUNNER)
+                trace.send_traces()
+                trace.prepare()
```
ranrib: Is this tab on purpose?

epsagon/runners/django.py
```diff
@@ -0,0 +1,69 @@
...
+            response.content
+        )
+
+        add_data_if_needed(
+            self.resource['metadata'],
+            'Response Headers',
+            dict(response.items())
+        )
+
+        if response.status_code >= 300:
```
ronyis: It should be 500 - fix also for flask.
(This is also what we do in node)

epsagon/wrappers/django.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+        self.runner = None
+        self.request = None
+        self.response = None
+
+        self.ignored_request = False
+
+    def __call__(self, request):
```
ronyis: There is no example on how to use the Django wrapping, I'm not sure why using `__init__` is not good enough as in flask (consider adding an example to the readme..)

epsagon/wrappers/django.py
```diff
@@ -0,0 +1,99 @@
...
+            return
+
+        # Ignoring non relevant content typoes.
+        if ignore_request(self.response.get('Content-Type').lower(), ''):
+            self.ignored_request = True
+            return
+
+        epsagon.trace.factory.get_trace().add_event(self.runner)
+        self.runner.update_response(self.response)
+        epsagon.trace.factory.get_trace().send_traces()
```
ronyis: Maybe you miss calling `prepare` at the end?
It happens in flask

epsagon/trace.py
```diff
@@ -46,28 +48,118 @@ def default(self, o):  # pylint: disable=method-hidden
...
+        if thread_id not in self.traces:
+            new_trace = Trace(
+                self.app_name,
+                self.token,
+                self.collector_url,
+                self.metadata_only,
+                self.disable_timeout_send,
+                self.debug
+            )
+            self.traces[thread_id] = new_trace
```
ronyis: I don't see that you ever remove the trace from this dict - it will cause a memory leak.
You should do that after attempting to send the trace, so `send_traces` should be moved to this class and it will always send the trace of the current thread.

epsagon/trace.py
```diff
@@ -113,8 +205,8 @@ def set_timeout_handler(self, context):
             if original_timeout <= TIMEOUT_GRACE_TIME_MS:
                 return
 
-            modified_timeout = (
-                original_timeout - TIMEOUT_GRACE_TIME_MS) / 1000.0
+            modified_timeout = \
```
ronyis: Basically using '\' in Python is a bad practice (https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

epsagon/trace.py
```diff
@@ -394,4 +485,4 @@ def send_traces(self):
 
 
 # pylint: disable=C0103
-tracer = Trace()
+factory = TraceFactory()
```
ronyis: Not very important, but I think that 'factory' is not a good name, especially when you import it directly so it's not clear that it is related to traces.

ranrib:
Please update README.MD with a short example, and another full example inside the examples/ dir

epsagon/runners/django.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, start_time, request):
+        """
+        Initialize.
+        :param start_time: event's start time (epoch).
+        :param request: the incoming request.
+        """
+        super(DjangoRunner, self).__init__(start_time)
+
+        self.event_id = str(uuid.uuid4())
+        self.resource['name'] = request.get_full_path()
```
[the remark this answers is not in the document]
LiadGoren: I have already discussed with @ranrib about it.

epsagon/trace.py
LiadGoren: Fixed

epsagon/trace.py
LiadGoren: Oops, didn't push it :|

epsagon/runners/django.py
LiadGoren: K

LiadGoren: Why 500 instead of 400?
I get why not 300, cause 300 are redirect codes

epsagon/modules/tornado.py
LiadGoren: nope

epsagon/runners/django.py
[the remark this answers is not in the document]
LiadGoren: @ranrib and I discussed about it...

epsagon/wrappers/django.py
LiadGoren: Can't see how it helps.
I think Flask did it because of multi-threading issues or w/e.
It also happens in the `before_request` method, so it's strange.

epsagon/wrappers/django.py
LiadGoren: Forgot about it, thanks.

epsagon/trace.py
LiadGoren: K

epsagon/trace.py
```diff
@@ -1,6 +1,7 @@
 """
 Trace object holds events and metadata
 """
+# pylint: disable=C0302
```
ronyis: Better to use names instead of these codes

epsagon/trace.py
```diff
@@ -461,6 +468,7 @@ def send_traces(self):
             return
         trace = ''
         try:
+            trace_factory.remove_trace()
```
ronyis: It's OK to do it in the beginning, but I think that it's better to do it in `finally` block - if some code will be added before this line it will be problematic (it could fail and the trace will not be removed)

epsagon/trace.py
```diff
@@ -114,6 +115,12 @@ def get_trace(self):
 
         return self.traces[thread_id]
 
+    def remove_trace(self):
```
ronyis: `remove_current_trace`?

epsagon/trace.py
```diff
@@ -112,30 +112,79 @@ def get_trace(self):
...
+    def add_exception(self, exception, stack_trace, additional_data=''):
+        """
+        add an exception to the relevant trace.
+        :param exception: the exception to add
+        :param stack_trace: the traceback at the moment of the event
+        :param additional_data: a json serializable object that contains
+            additional data regarding the exception
+        :return: None
+        """
+        if threading.currentThread().ident in self.traces:
```
ronyis: It was better not to duplicate this line - you could add a separate function and call it each time (even with a decorator :P)

epsagon/trace.py
LiadGoren: ok

epsagon/trace.py
LiadGoren: K :)

epsagon/trace.py
LiadGoren: K

epsagon/trace.py
LiadGoren: Yeah thought about it already lol :/

PR #80, opened 2019-05-12
fix(trace.py): support any object in trace encode

ranrib:
🆙 
also, applying the same trick to fix acceptance as it is in node.


acceptance/run.sh
```diff
@@ -12,9 +12,9 @@ function run_acceptance_test() {
     serverless remove --runtime ${runtime} --runtimeName ${runtimeName} --buildNumber ${build_num}
 }
 
-run_acceptance_test python2.7 python27
```
galbash: why

PR #83, opened 2019-05-15
fix(trace): support multiple tracers only for django

LiadGoren:


epsagon/trace.py
```diff
@@ -116,16 +141,20 @@ def get_or_create_trace(self):
...
         return self.traces.get(threading.currentThread().ident)
 
     def remove_current_trace(self):
         """
-        Remove the thread's trace.
+        Remove the thread's trace only if use_single_trace is set to False.
         """
-        self.traces.pop(threading.currentThread().ident, None)
+        if not self.use_single_trace:
+            self.traces.pop(threading.currentThread().ident, None)
```
ranrib: None is redundent

PR #95, opened 2019-06-04
feat(trace): support user-defined url blacklist

LiadGoren:


epsagon/wrappers/http_filters.py
```diff
@@ -52,6 +53,22 @@ def is_blacklisted_url(url):
...
 
+def is_payload_collection_blacklisted(url):
+    """
+    Return whether the payload should be collected according to the blacklisted
+    urls list in the Trace.
+    :param url: url string
+    :return:  True if URL is blacklisted, else False
+    """
+    url = urllib.parse.urlparse(url).netloc
+    trace_blacklist_urls = trace_factory.get_trace().url_patterns_to_ignore
```
galbash: 1. maybe make url_patterns_to_ignore a set on init (or an empty set if None was given so you don't have to worry about None value
2. then this function can become:
```
return any(blacklist_url in url for blacklist_url in trace_blacklist_urls)
```

epsagon/trace.py
```diff
@@ -102,6 +106,7 @@ def initialize(
         self.disable_timeout_send = disable_timeout_send
         self.debug = debug
         self.send_trace_only_on_error = send_trace_only_on_error
+        self.url_patterns_to_ignore = url_patterns_to_ignore
```
galbash:
```suggestion
        self.url_patterns_to_ignore = set(url_patterns_to_ignore) if url_patterns_to_ignore else set()
```

PR #96, opened 2019-06-05
fix(utils.py): set ignored urls via env

ranrib:
🆙

### DIFF
--- a/epsagon/utils.py
+++ b/epsagon/utils.py
@@ -80,6 +80,10 @@ def init(
         collector_url = get_tc_url(
             ((os.getenv('EPSAGON_SSL') or '').upper() == 'TRUE') | use_ssl
         )
+    # Ignored URLs is a comma separated values, if coming from env.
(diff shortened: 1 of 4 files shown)


epsagon/utils.py
```diff
@@ -80,6 +80,10 @@ def init(
         collector_url = get_tc_url(
             ((os.getenv('EPSAGON_SSL') or '').upper() == 'TRUE') | use_ssl
         )
+    # Ignored URLs is a comma separated values, if coming from env.
+    ignored_urls = os.getenv('EPSAGON_URLS_TO_IGNORE')
+    if ignored_urls:
+        ignored_urls = ignored_urls.split(',')
```
ronyis: Are the URLs are actually just hostnames?
If you are using `,` as delimiter the URLs must be encoded.
Anyway, I think that using a JSON list will be more convenient.

ranrib: Yes, it can be part of the hostname.
Why we need to encode `,`? it's not part of any hostname.
And at this point, I don't want to force users to set another configuration file..

PR #99, opened 2019-06-12
fix(constants.py): set timeout from env

ranrib:
🆙


epsagon/constants.py
```diff
@@ -15,6 +15,7 @@
 
 TIMEOUT_GRACE_TIME_MS = 200
 # How long we try to send traces in seconds.
-SEND_TIMEOUT = TIMEOUT_GRACE_TIME_MS / 1000.0
+TIMEOUT_ENV = float(os.getenv('EPSAGON_TIMEOUT', '0'))
```
ronyis: `EPSAGON_SEND_TIMEOUT_MS` is much better

ranrib: Your wish is my command

PR #100, opened 2019-06-13
fix(trace.py): fix strip operation

ranrib:
Also ignoring AWS requests from urllib3

### DIFF
--- a/epsagon/events/urllib3.py
+++ b/epsagon/events/urllib3.py
@@ -20,6 +20,9 @@ from ..wrappers.http_filters import (
 )
 from ..utils import update_api_gateway_headers
 
+# Ignore AWS endpoints here since we capture botocore.vendored.requests
+IGNORE_PATTERNS = ['.amazonaws.com']
+
 
 class Urllib3Event(BaseEvent):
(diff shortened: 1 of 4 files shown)


epsagon/trace.py
```diff
@@ -567,7 +567,8 @@ def _strip(self):
             if event.origin == 'runner' or event.origin == 'trigger':
                 continue
 
-            del self.events_map[event.identifier()]
+            if event.identifier() in self.events_map:
```
ronyis: Just know that you can do `self.events_map.pop(event.identifier(), None)`

epsagon/events/urllib3.py
```diff
@@ -20,6 +20,9 @@
 )
 from ..utils import update_api_gateway_headers
 
+# Ignore AWS endpoints here since we capture botocore.vendored.requests
+IGNORE_PATTERNS = ['.amazonaws.com']
```
ronyis: So `urllib3` instrumentation is used only by `botocore`?
Because the purpose was not to filter requests to API-GW.
I'm confused.

PR #101, opened 2019-06-16
fix(event.py): omit traceback using EPSAGON_OMIT_TRACEBACK

ranrib:
🆙


enoodle:
this is required by a customer

PR #102, opened 2019-06-16
fix(chalice.py): add Chalice specific wrapper

ranrib:
this needed since when we wrap the app, the user can't access to `app.attrs` (any attr).

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ app = Chalice(app_name="hello-world")
 def index():
     return {"hello": "world"}
 
-app = epsagon.lambda_wrapper(app)
+app = epsagon.chalice_wrapper(app)
 ```
 
 ### Zappa
--- a/epsagon/wrappers/chalice.py
+++ b/epsagon/wrappers/chalice.py
@@ -1,0 +1,31 @@
+"""
+Wrapper for AWS Lambda in Chalice environment.
+"""
+
+from __future__ import absolute_import
+from epsagon.wrappers import lambda_wrapper
+
+
+class ChaliceWrapper:
+    """
+    Class handles wrapping Chalice app.
+    In call we expect an invocation to come, and in getattr we allow `app.attr`
+    calls.
+    """
+    def __init__(self, app):
+        self.app = app
+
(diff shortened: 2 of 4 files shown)


epsagon/wrappers/chalice.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class ChaliceWrapper:
+    """
+    Class handles wrapping Chalice app.
+    In call we expect an invocation to come, and in getattr we allow `app.attr`
+    calls.
+    """
+    def __init__(self, app):
+        self.app = app
```
ronyis: You may want to use `_app` because it should not be changed (maybe add a property)

epsagon/wrappers/chalice.py
```diff
@@ -0,0 +1,31 @@
+"""
+Wrapper for AWS Lambda in Chalice environment.
+"""
+
+from __future__ import absolute_import
+from epsagon.wrappers import lambda_wrapper
+
+
+class ChaliceWrapper:
```
ronyis: Why not using inheritance?
More operators may be implemented, which are currently not wrapped (`__getitem__`, comparison, etc.).
Also, `isinstance` will keep working

README.md
```diff
@@ -195,7 +195,7 @@ app = Chalice(app_name="hello-world")
 def index():
     return {"hello": "world"}
 
-app = epsagon.lambda_wrapper(app)
+app = epsagon.chalice_wrapper(app)
```
ronyis: It looks like this line must be before defining the route functions - otherwise, I don't see how it could make any effect since `app` is no longer used.

ranrib: It's the opposite - user defines everything on the app, then we just wrap it for the invocation calls.

epsagon/wrappers/chalice.py
ranrib: 👍

epsagon/wrappers/chalice.py
ronyis: 👍

PR #103, opened 2019-07-07
Keys to ignore ep 1401

LiadGoren:


epsagon/trace.py
```diff
@@ -569,6 +578,34 @@ def _strip(self):
...
+
+    def remove_ignored_keys(self, input_dict):
+        """
+        Remove ignored keys recursively.
+        :param input_dict: Input dict to remove ignored keys from.
+        :return: None
+        """
+        if not self.keys_to_ignore:
+            return
+        stripped_keys = [self._strip_key(x) for x in self.keys_to_ignore]
```
ranrib: I would run this line once at the init, rather than everytime we send trace

LiadGoren: Got it

PR #104, opened 2019-07-08
feat(python_function): collect container metadata

LiadGoren:


epsagon/utils.py
```diff
@@ -137,3 +139,22 @@ def import_original_module():
...
+    Collects container metadata if exists.
+    :return: dict.
+    """
+    metadata_uri = os.environ.get('ECS_CONTAINER_METADATA_URI')
+    if not metadata_uri:
+        return {}
+    container_metadata = json.loads(requests.get(metadata_uri).content)
+
+    # Remove event from events list
+    events = list(trace_factory.get_trace().events_map.keys())
```
enoodle: I think we should remove only event with the correct type and uri (this might be an issue in the future if some other operation will be inserted before)

PR #108, opened 2019-07-17
feat(wrappers): use normalized http url as event name

tvaintrob:


epsagon/runners/flask.py
```diff
@@ -30,8 +30,8 @@ def __init__(self, start_time, app, request):
         self.event_id = str(uuid.uuid4())
 
         self.resource['name'] = (
-            ' '.join((app.name, request.url_rule.rule))
-            if request.url_rule
+            normalize_http_url(request.headers.get('Host', app.name))
```
ranrib: Running `normalize_http_url(app.name)` will yield a good result? (app.name is not a URL). I think maybe this one might be better:
```
normalize_http_url(request.headers['Host'])
if request.headers.get('Host')
else app.name
```

tvaintrob: normalize_http_url will fallback to the string itself if can’t parse it as a url so it will return the same result, still the proposed fix looks better

PR #116, opened 2019-08-06
fix(tornado.py): change finish wrapper

ranrib:
using `finish` instead of `on_finish`.
Also reporting errors only on >=500.
And setting the local version to be 0.0.0-dev (used only in local tests).

### DIFF
--- a/epsagon/constants.py
+++ b/epsagon/constants.py
@@ -2,7 +2,7 @@
 
 import os
 
-__version__ = '1.0.54'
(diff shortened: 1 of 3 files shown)


epsagon/constants.py
```diff
@@ -2,7 +2,7 @@
 
 import os
 
-__version__ = '1.0.54'
+__version__ = '0.0.0-dev'
```
galbash: I am pretty sure this will break the cicd

PR #117, opened 2019-08-07
fix(tornado.py): support concurrent requests in tornado

henprtz:


epsagon/trace.py
```diff
@@ -121,63 +121,80 @@ def switch_to_multiple_traces(self):
...
             self.traces[thread_id] = new_trace
         return self.traces[thread_id]
 
+    def get_trace_identifier(self):
+        """
+        Return the trace identifier
+        :return: trace identifier
+        """
+        return (
+            threading.currentThread().ident
```
ranrib: On Lambda this will be returned. Is that OK?

henprtz: Yes, Lambda runs in a single trace mode- This function is only called in case of multiple traces or when removing a trace with a unique id from the traces list.

PR #118, opened 2019-08-08
fix(http): allow custom error code

ranrib:
In addition:
1. Changed multiple tracers in Flask. This caused some other unrelated tests to break, since tracer_factory is global. Fixed.
2. Collect ECS task metadata in tornado.
3. Fixed task metadata event removal. which caused runner removal.

### DIFF
--- a/epsagon/modules/tornado.py
+++ b/epsagon/modules/tornado.py
@@ -9,6 +9,7 @@ import wrapt
 import epsagon.trace
 from epsagon.runners.tornado import TornadoRunner
 from epsagon.wrappers.http_filters import ignore_request
+from epsagon.utils import collect_container_metadata
 
 
 class TornadoWrapper(object):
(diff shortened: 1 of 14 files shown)


epsagon/modules/tornado.py
```diff
@@ -34,6 +37,15 @@ def before_request(cls, wrapped, instance, args, kwargs):
             if not ignored:
                 cls.RUNNER = TornadoRunner(time.time(), instance.request)
                 trace.set_runner(cls.RUNNER)
+
+                # Collect metadata in case this is a container.
+                if not cls.CHECKED_ECS_ENDPOINT:
```
galbash: just have the collect_container_metatdata() function cache it (can be a singleton) instead of caching it in every implementation

ranrib: Fixed

PR #131, opened 2019-08-19
Trace transport layer

tvaintrob:


epsagon/trace_transports.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class LogTransport(object):
+    """ send traces by logging them """
+
+    def __init__(self, token):
+        self.token = token
+
+    def send(self, trace):
+        trace_json = to_json({'token': self.token, 'trace': trace})
```
ranrib: We can extract the token from the trace itself, so no need to send it twice

epsagon/trace.py
```diff
@@ -719,6 +706,10 @@ def send_traces(self):
             return
         trace = ''
 
+        self.transport = self.transport \
```
ranrib: use `()` and not `\`

PR #139, opened 2019-08-29
Trace transport layer

tvaintrob:


epsagon/trace.py
```diff
@@ -28,6 +28,8 @@
     __version__
 )
 
+MAX_EVENTS_PER_TYPE = 20
```
ranrib: I think that these 2 lines revised accidentally

epsagon/trace_transports.py
```diff
@@ -23,7 +23,7 @@ class LogTransport(object):
 
     @classmethod
```
galbash: don't make it a class method for no reason. It can be a static method if you don't use self but the owner of this method is the instance. (Also we should have a united interface - it can't be a classmethod here but a regular method in the HTTPTransport)

galbash: also I would call it PrintTransport (since you use print and not logging) but doesn't matter

epsagon/utils.py
```diff
@@ -106,48 +107,57 @@ def init(
...
+            ),
+            url_patterns_to_ignore=ignored_urls or url_patterns_to_ignore,
+            keys_to_ignore=ignored_keys or keys_to_ignore,
+            transport=create_transport(collector_url, token))
+
+        # Append to ignored endpoints
+        http_filters.add_ignored_endpoints(ignored_paths or ignored_endpoints)
+    # pylint: disable=broad-except
+    except Exception:
+        logging.exception('exceptions occured')
```
galbash: are we sure we want this (especially the raise)? what errors can occur?

PR #140, opened 2019-09-01
feat(trace): split trace into multiple traces when it's too big to send

LiadGoren:


epsagon/trace.py
```diff
@@ -761,8 +768,47 @@ def remove_ignored_keys(self, input_dict):
...
+    def _send_trace_split(self):
+        """
+        Split trace into multiple traces and send them one after the other.
+        This is done by manipulating the trace object while keeping the same
+        runner.
+        :param trace: the trace to send.
+        """
+        # Get only events (without runner)
+        all_events = self.events.copy()
+        all_events.remove(self.runner)
```
henprtz: check that runner exists

epsagon/trace.py
```diff
@@ -761,8 +768,47 @@ def remove_ignored_keys(self, input_dict):
...
+        self.runner.resource['metadata']['fragment_seq'] = 1
+        self.clear_events()
+        self.add_event(self.runner)
+        for event in all_events:
+            self.events.append(event)
```
henprtz: why not use add_event?

epsagon/trace.py
```diff
@@ -761,8 +768,47 @@ def remove_ignored_keys(self, input_dict):
...
+            if self.length > self._max_trace_size:
+                self.events.pop()
```
henprtz: bug - you need to keep the popped event so we won't skip it.

epsagon/trace.py
```diff
@@ -761,8 +768,47 @@ def remove_ignored_keys(self, input_dict):
...
+        If trace size exceeds the maximum size, and split flag is on
+        then split the trace into multiple traces.
+        :return: None
+        """
+        if self.split_on_send and self.length > self._max_trace_size:
+            self._send_trace_split()
+        else:
+            self._send_traces()
+
+    def _send_trace_split(self):
```
henprtz: simplify this code by having an iterator that will return each time a batch of events that are <= trace size. and this way would also dumps only individual events instead of re-serializing the batch multiple time as it done here

for events_batch in get_next_events_batch(): events_batch.add(runner), increment fragment_id, send traces.

epsagon/trace.py
```diff
@@ -761,8 +768,47 @@ def remove_ignored_keys(self, input_dict):
...
+        """
+        Split trace into multiple traces and send them one after the other.
+        This is done by manipulating the trace object while keeping the same
+        runner.
+        :param trace: the trace to send.
+        """
+        # Get only events (without runner)
+        all_events = self.events.copy()
```
henprtz: deepcopy?

epsagon/trace.py
```diff
@@ -439,7 +447,7 @@ def set_timeout_handler(self, context):
                 return
 
             modified_timeout = (
-                original_timeout - TIMEOUT_GRACE_TIME_MS) / 1000.0
+                           original_timeout - TIMEOUT_GRACE_TIME_MS) / 1000.0
```
henprtz: ?

henprtz:
Add UT please @LiadGoren

epsagon/trace.py
LiadGoren: Indentation shit

epsagon/trace.py
LiadGoren: Yep

epsagon/trace.py
LiadGoren: You are right :)

epsagon/trace.py
LiadGoren: I don't need to deep copy the events objects. only the list (so it won't be affected when looping )

epsagon/trace.py
LiadGoren: You have to dump the events to JSON in order to understand whether it's size to too big.
It'll be fixed in a later version of our instrumentation code (as discussed with @ranrib ).

PR #168, opened 2019-11-19
test(acceptance): adding python 3.8 acceptance tests

galbash:


acceptance/run.sh
```diff
@@ -12,9 +13,7 @@ function run_acceptance_test() {
     serverless remove --runtime ${runtime} --runtimeName ${runtimeName} --buildNumber ${build_num}
 }
 
-run_acceptance_test python2.7 p27
-run_acceptance_test python3.6 p36
-#run_acceptance_test python3.7 p37
+run_acceptance_test python${version} py${version//.}
```
henprtz: py? not p ?

.travis.yml
```diff
@@ -11,14 +14,23 @@ install:
 
 script:
   - "pylint --msg-template='{path}:{line}: [{msg_id}({symbol}) {obj}] {msg}' epsagon/"
-  - tox
+  - pytest
```
henprtz: explain?

.travis.yml
```diff
@@ -11,14 +14,23 @@ install:
...
   - npm install && export PATH=$(pwd)/node_modules/.bin:$PATH
-  - ./acceptance/run.sh $TRAVIS_BUILD_NUMBER
+  - ./acceptance/run.sh $TRAVIS_BUILD_NUMBER $TRAVIS_PYTHON_VERSION
 
+jobs:
+  include:
+    - stage: build-and-deploy
+      provider: script
+      python:
+        - "3.8"
```
ronyis: What's that used for?
It doesn't override the global Python version?

galbash: yes, only for deploy

.travis.yml
galbash: no need, we now have a separate machine for each build

acceptance/run.sh
galbash: works idk

PR #175, opened 2019-12-24
feat(lambda): collecting status code from result

galbash:


epsagon/event.py
```diff
@@ -105,3 +107,5 @@ def set_exception(self, exception, traceback_data):
         self.exception['message'] = str(exception)
         self.exception['traceback'] = traceback_data
         self.exception['time'] = time.time()
+        if not handled:
```
ranrib: Wouldn't it be a bit more clear to call it "manually_set" instead of handled/unhandled?

ronyis: I think that 'handled' it's good. It can be 'handled' and not manually set - for example, if we will extract errors from calling `logging.error`

epsagon/event.py
```diff
@@ -105,3 +107,5 @@ def set_exception(self, exception, traceback_data):
...
+            self.exception.setdefault('additional_data', {})['unhandled'] = True
```
ronyis: I think it's more straight forward to save `{'handled': True}` so that the trace will be not changed in the most common event of unhandled exception.

ronyis: Are you sure that `additional_data` is needed?

epsagon/events/botocore.py
```diff
@@ -64,14 +64,15 @@ def __init__(self, wrapped, instance, args, kwargs, start_time, response,
...
-        :param exception: Exception object
-        :param traceback_data: traceback string
-        :return: None
+        see {Event.set_exception}
         """
-        super(BotocoreEvent, self).set_exception(exception, traceback_data)
+        super(BotocoreEvent, self).set_exception(
+            exception,
+            traceback_data,
+            handled
```
ronyis: Keyword args should be called with the keyword.

epsagon/wrappers/aws_lambda.py
```diff
@@ -94,10 +108,11 @@ def _lambda_wrapper(*args, **kwargs):
             return result
         # pylint: disable=W0703
         except Exception as exception:
-            runner.set_exception(exception, traceback.format_exc())
+            runner.set_exception(exception, traceback.format_exc(), False)
```
ronyis: Call with keyword (in all places)

epsagon/wrappers/aws_lambda.py
```diff
@@ -22,6 +23,19 @@
...
 
+def _add_status_code(runner, return_value):
+    """
+    Tries to extract the status code from the return value and add it
+    as a metadata field
+    :param runner: Runner event to update
+    :param return_value: The return value to extract from
+    """
+    if isinstance(return_value, collections.Mapping):
+        status_code = return_value.get('statusCode', None)
```
ronyis: `None` is not needed

epsagon/wrappers/aws_lambda.py
```diff
@@ -94,10 +108,11 @@ def _lambda_wrapper(*args, **kwargs):
             return result
         # pylint: disable=W0703
         except Exception as exception:
-            runner.set_exception(exception, traceback.format_exc())
+            runner.set_exception(exception, traceback.format_exc(), False)
             raise
         finally:
             try:
+                _add_status_code(runner, result)
```
ronyis: It would be better if you collect it only for API-GW / load balancer trigger, and I think you should save it in the trigger.
(Now you could mistakenly collect this data when it's `metadata_only=True`)

epsagon/wrappers/aws_lambda.py
```diff
@@ -206,10 +221,11 @@ def _lambda_wrapper(*args, **kwargs):
             return result
         # pylint: disable=W0703
         except Exception as exception:
-            runner.set_exception(exception, traceback.format_exc())
+            runner.set_exception(exception, traceback.format_exc(), False)
```
ronyis: Code duplication is so fun! 🎉

epsagon/wrappers/aws_lambda.py
```diff
@@ -206,10 +221,11 @@ def _lambda_wrapper(*args, **kwargs):
...
             raise
         finally:
             try:
+                _add_status_code(runner, result)
```
ronyis: Step functions can be triggered by HTTP event?

tests/test_trace.py
```diff
@@ -861,7 +861,7 @@ def test_event_with_datetime(wrapped_post):
 def test_send_on_error_only_off_with_error(wrapped_post):
     trace = trace_factory.get_or_create_trace()
     trace.token = 'a'
-    trace.runner = mock.MagicMock()
+    trace.runner = RunnerEventMock()
```
ronyis: This is not really related to your PR or I missed something?

epsagon/wrappers/aws_lambda.py
galbash: we agreed to allow it

epsagon/wrappers/aws_lambda.py
galbash: :)

galbash: this is why node is better 😎

epsagon/wrappers/aws_lambda.py
galbash: actually i think it can :)

tests/test_trace.py
galbash: fixing the random crashes in the tests

ronyis: 👑

PR #180, opened 2020-01-07
Fixing trace stripped size calculation

maorlx:
Getting the correct stripped data size at trace stripping


ranrib:
@maorlx  merge or close?

PR #186, opened 2020-01-16
fix(metadata_trimming): trim all metadata fields

maorlx:


tests/test_trace.py
```diff
@@ -511,6 +520,27 @@ def test_send_big_trace(wrapped_post):
         headers={'Authorization': 'Bearer {}'.format(trace.token)}
     )
```
razlo: let's add a test for this specific case - 
have a key to ignore in return value, and validate it's actually removed

PR #190, opened 2020-01-22
Ignored keys removal ep 2228

maorlx:


epsagon/trace.py
```diff
@@ -850,22 +850,38 @@ def _strip_key(key):
...
+        # pylint: disable=too-many-nested-blocks
+        copied_dict = None
         if self.keys_to_ignore:
             # Python 2 returns a list, while Python3 returns an iterator.
-            for key, value in list(input_dict.items()):
+            for key in input_dict:
                 if self._strip_key(key) in self.keys_to_ignore:
-                    input_dict.pop(key)
+                    if copied_dict is None:
+                        copied_dict = input_dict.copy()
```
razlo: I think there's no point in actually shallow copying the data.
let's just iterate through the previous, and filter out the key we don't want

epsagon/trace.py
```diff
@@ -850,22 +850,38 @@ def _strip_key(key):
...
                     if isinstance(value, dict):
-                        self.remove_ignored_keys(value)
+                        has_changed, result = self.remove_ignored_keys(value)
+                        if has_changed:
+                            if copied_dict is None:
+                                copied_dict = input_dict.copy()
+                            copied_dict[key] = result
+
+        has_changed = copied_dict is not None
+        return has_changed, copied_dict if has_changed else input_dict
```
razlo: also no need for this has_changed flag.
assign the result of this function straight to `event.resource[metadata]`.
If nothing was removed, the original dict will be returned

epsagon/trace.py
```diff
@@ -850,22 +850,38 @@ def _strip_key(key):
 
     def remove_ignored_keys(self, input_dict):
         """
-        Remove ignored keys recursively.
+        Remove ignored keys recursively in input_dict.
+        If an ignored key has been found in a dict, then the
+        dict is copied (shallow copy) and the ignored key is removed.
         :param input_dict: Input dict to remove ignored keys from.
-        :return: None
+        :return: tuple of -
```
razlo: should just return a dict. see later comment

PR #194, opened 2020-02-08
feat(tracer): propagate identifier flag

henprtz:


ranrib:
Let's use a less scary and more meaningful term for this please - something like "ADD_ID_TO_RESPONSE"

epsagon/constants.py
```diff
@@ -56,3 +56,4 @@ def is_strong_key(key):
 
 
 STEP_DICT_NAME = 'Epsagon'
+TRACE_ID_DICT_NAME = '_epsagon_trace_id'
```
ronyis: This name is not good. It's no the trace ID and it's not a dict.
Maybe just call it `epsagon_event_id`?

epsagon/utils.py
```diff
@@ -109,6 +110,8 @@ def init(
       collection.
     :param keys_to_ignore: List of keys to ignore while extracting metadata.
     :param ignored_endpoints: List of ignored endpoints for web frameworks.
+    :param split_on_send: Split the trace on send flag
+    :param propagate_id: Inject identifiers via return value flag
```
ronyis: It's relevant only for Lambda, and I don't see that it gonna change, maybe update the name? (Also the env var name?)

epsagon/trace.py
```diff
@@ -108,6 +109,7 @@ def initialize(
             keys_to_ignore,
             transport,
             split_on_send,
+            propagate_id,
```
ronyis: Update doc?

tests/wrappers/test_lambda_wrapper.py
```diff
@@ -561,3 +561,84 @@ def wrapped_lambda(event, context):
...
+    side_effect=['trigger']
+)
+def test_propagate_id_to_dict_sanity(
+        trigger_factory_mock,
+        _,
+        set_exception_mock
+):
+    retval = {
+        'hello': 2,
+        '_epsagon_trace_id': 'test_request_id'
```
ronyis: Better use const here

PR #196, opened 2020-02-10
fix(trace.py): add support for strings in set_error

ranrib:


README.md
```diff
@@ -155,22 +155,28 @@ main()
...
-  pass
+    epsagon.label('key', 'value')
+    epsagon.label('user_id', event['headers']['auth'])
+    epsagon.label('number_of_records_parsed_successfully', 42)
 ```
 
 ### Custom Errors
 
-Set a custom error, maybe without even failing the function:
+Set a custom error, even if handled correctly:
```
razlo: I think we need a bit more verbose documentation here.
Right now the docs don't mention anything about how we aggregate issues
(all different "string" set_errors will be aggregated to the same issue as far as I remember)

epsagon/trace.py
```diff
@@ -719,6 +719,9 @@ def set_error(self, exception, traceback_data=None):
                 traceback_data = ''.join(
                     traceback.format_list(traceback.extract_stack())
                 )
+        # Convert exception string to Exception type
+        if isinstance(exception, str):
+            exception = Exception(exception)
```
razlo: Maybe let's define an EpsagonException so user defined set_error(Exception()) won't be aggregated with "string" errors?

not a must

ranrib: Done, good call

README.md
ranrib: I agree, I put a reference here, because it is not the right place to document product behaviour

PR #202, opened 2020-02-23
feat(stepfunctions): Improve aws stepfunctions support

idonava:
Add support to the following functions: send_task_success, send_task_heartbeat, describe_execution
Add support to step function lambda triggered by SQS

### DIFF
--- a/epsagon/events/botocore.py
+++ b/epsagon/events/botocore.py
@@ -12,7 +12,7 @@ import simplejson as json
 from epsagon.constants import STEP_DICT_NAME
 from ..trace import trace_factory
 from ..event import BaseEvent
-from ..utils import add_data_if_needed, add_data_to_resource_metadata
+from ..utils import add_data_if_needed, add_metadata_from_dict
 
 # Conditionally importing boto3
 ClientError = Exception  # pylint: disable=invalid-name
@@ -1393,7 +1393,7 @@ class BotocoreStepFunctionEvent(BotocoreEvent):
         :return: None
         """
         _, request_args = args
-        add_data_to_resource_metadata(self.resource, request_args, 'taskToken')
+        add_metadata_from_dict(self.resource, request_args, 'taskToken')
 
     def process_describe_execution_operation(self, args, _):
         """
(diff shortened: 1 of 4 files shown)


epsagon/events/botocore.py
```diff
@@ -1380,6 +1401,34 @@ def process_start_exec_response(self, response):
...
+            ''
+        )
+
+    def process_describe_execution_response(self, response):
+        """
+        Process describeExecution response
+        :param response: response from Step function Client
+        :return: None
+        """
+        self.resource['metadata']['Execution ARN'] = response.get(
```
henprtz: what happens if this is None?

epsagon/events/botocore.py
```diff
@@ -1380,6 +1401,34 @@ def process_start_exec_response(self, response):
...
+            'executionArn')
+        self.resource['metadata']['Input'] = response.get('input')
+        self.resource['metadata']['Name'] = response.get('name')
+        self.resource['metadata']['Output'] = response.get('output')
+        self.resource['metadata']['Start Date'] = response.get('startDate')
+        self.resource['metadata']['Status'] = response.get('status')
+        stop_date = response.get('stopDate')
+        if stop_date:
+            self.resource['metadata']['Stop Date'] = stop_date
+
```
henprtz: why are you only checking if stop date exists?

epsagon/utils.py
```diff
@@ -286,3 +290,45 @@ def get_traceback_data_from_exception(exception):
...
+    :return: Title Case string
+    """
+    if not isinstance(camel_case_string, str):
+        return None
+    title_case = re.sub('([^-])([A-Z][a-z-]+)', r'\1 \2', camel_case_string)\
+        .title()
+    return title_case
+
+
+def add_data_to_resource_metadata(resource, dictionary, key):
```
ronyis: *`add_metadata_from_dict` ?

epsagon/utils.py
```diff
@@ -286,3 +290,45 @@ def get_traceback_data_from_exception(exception):
...
+
+def parse_json(json_string):
+    """
+    Parse JSON string to a Python Dictionary
+    :param json_string: JSON string
+    :return: Python Dictionary
+    """
+    try:
+        return json.loads(json_string)
+    except ValueError:
```
ronyis: That's not the exception type when JSON load fails (check it)

PR #208, opened 2020-03-01
feat(main): run epsagon using env var

ranrib:
In additon - auto-tracing Flask and Django apps.


epsagon/__init__.py
```diff
@@ -13,6 +13,14 @@
     from .handler import wrapper
 
 
+def auto_load(_):
```
ranrib: Check when entry point is being used

ranrib: check init twice

ranrib: check when `AUTOWRAPT_BOOTSTRAP=epsagon` called but no Epsagon installed

ranrib: Lock wrapt and autowrapt versions

ranrib: Make sure we don't auto-trace Tornado twice

ranrib: 1. Entrypoints are exposing an interface to interact with the code. They are not being called automatically.
2. If init is being called twice, the later init will take place. If environment variables are supplied they will be the one to be chosen. There should be any problem raised from any relevant scenario.
3. If Epsagon is not installed but `AUTOWRAPT_BOOTSTRAP=epsagon` is called - nothing happens.
4. versions locked
5. Tornado it being traced in a different way (instrumenting incoming call, and not adding a middleware to the app) - so it can't be done twice

razlo: I don't see that the versions are locked
forgot to push?

PR #210, opened 2020-03-02
feat(tracer): min status code error flag.

henprtz:


epsagon/trace.py
```diff
@@ -939,6 +946,16 @@ def _send_traces(self):
         """
         if self.token == '' or self.trace_sent:
             return
+        if self.min_error_code and self.runner:
```
ranrib: Move all this logic into a different function (and please make sure it is fail-safe)

epsagon/trace.py
```diff
@@ -144,6 +146,7 @@ def initialize(
         self.transport = transport
         self.split_on_send = split_on_send
         self.propagate_lambda_id = propagate_lambda_id
+        self.min_error_code = min_error_code
```
ranrib: What about env var? And also update the readme please.

epsagon/trace.py
ranrib: Also please add a small UT for that

ranrib:
@henprtz deploy or close?

ranrib:
@henprtz closed

PR #220, opened 2020-04-23
Add trace id to logs

enoodle:


epsagon/modules/logging.py
```diff
@@ -24,10 +28,41 @@ def _wrapper(wrapped, _instance, args, kwargs):
     return wrapped(*args, **kwargs)
 
 
+def _get_log_id():
+    """
+    retrieve or generate the log id for this trace
+    """
+    return trace_factory.get_log_id()
```
razlo, line 35: let's not generate this log id for Lambda and StepLambda runners, since we don't need them there
I'm not sure where to do it in the instrumentation actually - let's think about it

epsagon/modules/logging.py
```diff
@@ -3,11 +3,15 @@
 """
 
 from __future__ import absolute_import
+from functools import partial
 import os
 import wrapt
 from ..trace import trace_factory
 
 
+LOG_ID_LABEL_NAME = 'epsagon_log_id'
```
razlo, line 12: unused?

tests/test_trace.py
```diff
@@ -1309,7 +1321,8 @@ def test_init_propagate_lambda_identifier_env(wrapped_init, _create):
         keys_to_ignore=None,
         keys_to_allow=None,
         split_on_send=False,
-        propagate_lambda_id=True
+        propagate_lambda_id=True,
+        add_log_id=False,
```
razlo, line 1325: let's also add an environment variable to control it, and also add it to the env vars table in the README

epsagon/modules/logging.py
```diff
@@ -24,10 +28,41 @@ def _wrapper(wrapped, _instance, args, kwargs):
...
     """
     wrapt.wrap_function_wrapper('logging', 'exception', _wrapper)
     wrapt.wrap_function_wrapper('logging', 'Logger.exception', _wrapper)
+    wrapt.wrap_function_wrapper(
+        'logging', 'Logger.log', partial(_epsagon_log_id_wrapper, 1))
+    wrapt.wrap_function_wrapper(
+        'builtins', 'print', partial(_epsagon_log_id_wrapper, 0))
+    for log_function in ['info', 'debug', 'error',
+                         'warning', 'exception', 'critical']:
+        wrapt.wrap_function_wrapper(
```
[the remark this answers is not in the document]
enoodle: No, it will call the Logger function to implement it and will cause us to add the log id twice

epsagon/trace.py
```diff
@@ -382,6 +389,20 @@ def add_label(self, key, value):
...
+        """
+        Get the value of the add_log_id flag
+        """
+        return self.add_log_id
+
+    def get_log_id(self):
+        """
+        Get the log id of the current trace
+        """
+        if self.get_trace():
```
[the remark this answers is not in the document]
enoodle: more clear and simple this way.. it is just getting an attribute no doing something heavy

epsagon/trace.py
```diff
@@ -725,6 +749,17 @@ def add_label(self, key, value):
             return
         self.custom_labels[key] = value
 
+    def get_log_id(self):
+        """
+        Get the log id if the add_log_id flag is on, else return None
+        """
+        if self.add_log_id:
+            if self.log_id:
```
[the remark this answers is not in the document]
enoodle: It is not always set to true, so sometimes it is empty anyway.

epsagon/modules/logging.py
enoodle: It will only generate it if the flag is set to true, which will not happen for lambdas..

epsagon/modules/logging.py
enoodle: :+1:

tests/test_trace.py
enoodle: :+1:

epsagon/trace.py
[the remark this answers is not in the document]
maorlx: I mean only if add_log_id is true

epsagon/trace.py
```diff
@@ -1012,6 +1047,9 @@ def _send_traces(self):
             if self.runner:
                 self.runner.terminate()
 
+                if self.log_id:
+                    self.runner.resource['metadata']['log_id'] = self.log_id
```
razlo, line 1051: we need to put it not under 'metadata', it's an internal field of ours

epsagon/modules/logging.py
```diff
@@ -24,10 +25,41 @@ def _wrapper(wrapped, _instance, args, kwargs):
     return wrapped(*args, **kwargs)
 
 
+def _get_log_id():
```
razlo, line 28: @galbash @ronyis 
what was decided about log_id / trace_id?

epsagon/trace.py
enoodle: I dont under stand, this is `metadata`

PR #222, opened 2020-04-25
feat(trace.py): add support for numeric labels

ranrib:


epsagon/trace.py
```diff
@@ -685,23 +685,26 @@ def verify_custom_label(self, key, value):
...
-                len(value) +
+                len(str(value)) +
                 self.custom_labels_size > MAX_LABEL_SIZE
         ):
             return False
 
-        self.custom_labels_size += len(key) + len(value)
+        self.custom_labels_size += len(key) + len(str(value))
 
         return True
```
ronyis, line 710: Need to update `add_label` doc.
You should mention what types are expected, and the dict flattening.

epsagon/trace.py
```diff
@@ -717,9 +720,9 @@ def add_label(self, key, value):
                 self.add_label('{}.{}'.format(key, dict_key), dict_value)
             return
 
-        # Convert numbers to string.
-        if isinstance(value, (int, float)):
-            value = str(value)
+        # Convert boolean to string.
+        if isinstance(value, bool):
```
ronyis, line 724: `bool` should be saved as is (it's a valid JSON type).
Note that it's a subtype of `int`.

epsagon/trace.py
```diff
@@ -685,23 +685,26 @@ def verify_custom_label(self, key, value):
...
+            print('EPSAGON: label key support only string type')
+            print('Received {key}:{value}'.format(key=key, value=value))
+            return False
+        if not isinstance(value, (int, float, str)):
+            print('EPSAGON: label value support only string, int, float types')
             print('Received {key}:{value}'.format(key=key, value=value))
             return False
 
-        if len(key) + len(value) > MAX_LABEL_SIZE:
+        if len(key) + len(str(value)) > MAX_LABEL_SIZE:
```
ronyis, line 697: I think it's correct but tricky - only because we do a JSON dump of the labels (of the whole trace actually).
A comment will be nice.

ronyis, line 697: I just looked and `MAX_LABEL_SIZE` is actually greater than `MAX_TRACE_SIZE_BYTES` (fun).

epsagon/trace.py
```diff
@@ -685,23 +685,26 @@ def verify_custom_label(self, key, value):
...
             return False
 
         if (
                 len(key) +
-                len(value) +
+                len(str(value)) +
                 self.custom_labels_size > MAX_LABEL_SIZE
```
ronyis, line 703: If you have this check, the previous check is redundant...

epsagon/trace.py
```diff
@@ -685,23 +685,26 @@ def verify_custom_label(self, key, value):
         :param value: Value for the label data (string)
         :return: True/False
         """
-        if not isinstance(key, str) or not isinstance(value, str):
-            print('EPSAGON: epsagon.label() only supports '
-                  'label keys and values of type string.')
+        if not isinstance(key, str):
+            print('EPSAGON: label key support only string type')
+            print('Received {key}:{value}'.format(key=key, value=value))
```
ronyis, line 690: I would just print the type of the key, in the same line (definitely not the value which can be very big).

epsagon/trace.py
ranrib: 👍

epsagon/trace.py
ranrib: 👍

epsagon/trace.py
ranrib: 👍

ranrib: Reduced to 10

epsagon/trace.py
ranrib: math

epsagon/trace.py
ranrib: 👍

ranrib:
Please re-review @ronyis

PR #225, opened 2020-05-03
Collect k8s container

enoodle:


ranrib:
@enoodle make sure to do semantic commit when you squash and merge

enoodle:
@ranrib the first 4 commits are semantic, it will use them.

PR #234, opened 2020-05-17
feat(aws_lambda.py): support custom apth for steps dict

ranrib:


epsagon/trace.py
```diff
@@ -1087,7 +1095,7 @@ def _send_traces(self):
             ))
         finally:
             if self.debug:
-                pprint.pprint(self.to_dict())
+                print('trace:', trace)
```
ranrib, line 1098: This fixes the trace print to be in one line and readable, instead of many lines.

epsagon/utils.py
```diff
@@ -249,32 +259,36 @@ def collect_container_metadata(metadata):
...
     recursively search for a key in an object
     :param obj: The dict to search in
     :param key: The key to search for
-    :return: The value of this key in the object, or None if was not found
+    :param path: The path (as an array) of the key nested in the object
+    :return: The value and path of this key in the object,
+        or None if was not found
     """
-    result = None
+    if not path:
```
razlo, line 271: just use [] as default?

PR #241, opened 2020-06-15
[EP-3515] Support Gunicorn

henprtz:


ranrib:
Please make sure to add an appropriate logo + update the docs and README

henprtz:
@ranrib - Fixed and added README.

Icon & Documentation are in different PRs (plus, you LOVE adding icons)

epsagon/modules/gunicorn.py
```diff
@@ -0,0 +1,191 @@
...
+            runner = GunicornRunner(
+                time.time(),
+                req
+            )
+            trace.set_runner(runner)
+
+            # Collect metadata in case this is a container.
+            collect_container_metadata(runner.resource['metadata'])
+        # pylint: disable=W0703
+        except Exception as exception:
```
adavidai, line 142: Looks like an Epsagon error will be created for instrumentation failure as well as failure in the original gunicorn code. Can we separate those? (As one is something the customer is interested in, and the other is internal and we want to know about it)

epsagon/modules/gunicorn.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def __init__(self, start_time, request):
+        """
+        Initialize.
+        :param start_time: event's start time (epoch).
+        :param request: the incoming request.
+        """
+        super(GunicornRunner, self).__init__(start_time)
+
+        self.event_id = str(uuid.uuid4())
```
adavidai, line 74: Starting this point this looks like a general http based code (getting the hostname, the path and the headers). Can we inherit an HttpEvent instead of BaseEvent? (Extract mutual code for http instrumentations)

epsagon/modules/gunicorn.py
henprtz: That's not true. These code only runs as part of a server hook, meaning it has zero effect on the customers code.

But thank you for mentioning that it made me realize I'm not catching the customer's code exception due to the design of those server hooks.

There's a bug in the gunicorn library that exception are not being propagated to the server hooks, I'm going to have to find a different method to hook.

epsagon/modules/gunicorn.py
henprtz: We can't. The data looks different than a generic HTTP request and response object as HttpEvent expects it to be, it didn't work.

henprtz:
Release notes: This version doesn't support exceptions during requests. The trace will appear with 200. This is a known issue due to the limitation under gunicorn's server hooks.

PR #243, opened 2020-06-17
Revert "fix(logging.py): fix default logging exception capture behavior"

ronyis:
Reverts epsagon/epsagon-python#242


ronyis:
This is a temp workaround for a bug when capturing loggings into trace errors

PR #245, opened 2020-06-17
fix(logging.py): fix default logging exception capture behavior

ranrib:


tests/test_logging.py
```diff
@@ -0,0 +1,40 @@
...
+def test_logging_exception_capture_with_args():
+    retval = 'success'
+
+    @epsagon.wrappers.python_function.python_wrapper
+    def wrapped_function(event, context):
+        logging.exception('test %s %s', 'test', 'test')
+        return retval
+
+    assert wrapped_function('a', 'b') == retval
+    trace_mock.set_error.assert_called_with('test %s %s', None)
```
adavidai, line 40: Can we have "test test test" as the error in Epsagon (instead of "test %s %s")?

PR #246, opened 2020-06-25
feat(azure.py): add support for CosmosDB

ranrib:


epsagon/events/azure.py
```diff
@@ -0,0 +1,227 @@
...
+        self.OPERATION_TO_FUNC.update({
+            'upsert_item': self.upsert_item_request,
+            'delete_item': self.delete_item_request,
+            'query_items': self.query_items_request,
+        })
+        self.kwargs = kwargs
+        self.args = args
+        self.response = response
+
+        self.resource['metadata']['azure.location'] = (
```
adavidai, line 144: How solid is this format? consider try except, in case there are variants we didn't account for yet (e.g. no "_" so [1] will fail)

epsagon/events/azure.py
```diff
@@ -0,0 +1,227 @@
...
+            instance.client_connection.ReadEndpoint.split(
+                '-'
+            )[1].split('.')[0]
+        )
+        self.resource['metadata']['azure.cosmos.endpoint'] = (
+            instance.client_connection.url_connection
+        )
+        _, database, _, container = instance.container_link.split('/')
```
adavidai, line 152: Does container name have its tag as well (e.g. `:latest`)?

epsagon/events/azure.py
```diff
@@ -0,0 +1,227 @@
...
+        self.resource['name'] = container
+        self.resource['metadata']['azure.cosmos.database_id'] = database
+        self.resource['metadata']['azure.cosmos.container_id'] = container
+
+        self.OPERATION_TO_FUNC.get(self.resource['operation'], empty_func)()
+
+    def upsert_item_request(self):
+        self.resource['metadata']['azure.cosmos.document_id'] = (
+            self.args[0].get('id')
```
adavidai, line 161: Is there always an `id`? if not maybe don't add it to metadata?
(same for delete item operation, and query on query items operation)

epsagon/events/azure.py
ranrib: container in this case = a dynamodb table. names are a bit confusing at the beginning

epsagon/events/azure.py
ranrib: The instrumentation in Azure case is very specific (this looks only on container operations) - so this field remains in the same structure

PR #252, opened 2020-07-13
feat(python_wrapper): Improving generic Python function wrapper support

ronyis:


epsagon/wrappers/python_function.py
```diff
@@ -52,8 +54,7 @@ def wrap_python_function(func, args, kwargs):
         raise
     finally:
         try:
-            if not epsagon.trace.trace_factory.get_trace().metadata_only:
-                runner.resource['metadata']['return_value'] = result
+            runner.add_json_field('python.function.return_value', result)
```
adavidai, line 57: Does changing the way the return value is sent (the name of the tag) requires adaptations in the backend as well?

epsagon/wrappers/python_function.py
```diff
@@ -67,12 +68,41 @@ def wrap_python_function(func, args, kwargs):
...
+    -   @python_wrapper()
+        def my_function(params):
+            ...
+
+    -   @python_wrapper
+        def my_function(params):
+            ...
+
+    NOTE: Should not be used for two functions in the same stack trace, or when
+          running under a traced context (e.g. AWS Lambda, HTTP frameworks, etc)
```
adavidai, line 92: What happens if it IS used inside already traced code? (adding some noise to the graph in Epsagon is ok, as long as it has no implications on the instrumented code)

epsagon/runners/python_function.py
```diff
@@ -37,21 +39,36 @@ def __init__(
...
         self.event_id = str(uuid.uuid4())
-        self.resource['name'] = wrapped_function.__name__
+        self.resource['name'] = name if name else wrapped_function.__name__
         self.resource['operation'] = self.OPERATION
+
         self.resource['metadata'].update({
-            'module': wrapped_function.__module__,
-            'args_length': len(wrapped_args),
-            'kwargs_length': len(wrapped_kwargs)
+            'python.module': wrapped_function.__module__,
```
adavidai, line 46: Does this renaming of the metadata require any adaptation in the Epsagon backend as well?

epsagon/wrappers/python_function.py
```diff
@@ -67,12 +68,41 @@ def wrap_python_function(func, args, kwargs):
...
+    """
+    name = kwargs.get('name')
+
+    def _inner_wrapper(func):
+
+        @functools.wraps(func)
+        def _python_wrapper(*args, **kwargs):
+            epsagon.trace.trace_factory.get_or_create_trace().prepare()
+            return wrap_python_function(func, args, kwargs, name=name)
```
adavidai, line 101: Was this a bug before? Weird...

tests/wrappers/test_python_function.py
```diff
@@ -26,7 +26,7 @@ def setup_function(func):
 def test_function_wrapper_sanity(_, ):
     retval = 'success'
 
-    @epsagon.wrappers.python_function.python_wrapper
+    @epsagon.wrappers.python_function.python_wrapper(name='test-func')
```
adavidai, line 29: I think it is better to add a test (duplicate this one, share relevant code) for python_wrapper with name. That way we still have a test for without a name and see that it didn't break

examples/custom_labels.py
```diff
@@ -6,17 +6,13 @@
...
 @epsagon.lambda_wrapper
-def handle():
+def handle(event, context):
     epsagon.label('label', 'something_to_filter_afterwards')
     epsagon.label('number_of_records_parsed_successfully', 42)
 
     return 'It worked!'
-
-
-if __name__ == '__main__':
```
adavidai, line 21: Why not leave the main for this example, to allow running it using `python custom_labels.py` ?

epsagon/wrappers/python_function.py
ronyis: Nope

epsagon/wrappers/python_function.py
ronyis: The runner will be overridden - though other previous events will remain.

Not so good, but will not break.

epsagon/runners/python_function.py
ronyis: Nope

epsagon/wrappers/python_function.py
ronyis: No, this change is needed to support using the decorator both as `@python_wrapper` (for backward compatibility) and `@python_wrapper()`

tests/wrappers/test_python_function.py
ronyis: We have other tests that don't use a name. Do you think another one is still needed?

examples/custom_labels.py
ronyis: Because it's a lambda handler in this example, and mostly it doesn't make sense to run it like that (so I think it's just confusing).

PR #258, opened 2020-07-23
Working on transport fixture to UT

ronyis:
Not ready yet

### DIFF
--- a/epsagon/trace.py
+++ b/epsagon/trace.py
@@ -211,11 +211,12 @@ class TraceFactory(object):
             url_patterns_to_ignore=self.url_patterns_to_ignore,
             keys_to_ignore=self.keys_to_ignore,
             keys_to_allow=self.keys_to_allow,
-            unique_id=unique_id,
+            transport=self.transport,
             split_on_send=self.split_on_send,
             propagate_lambda_id=self.propagate_lambda_id,
             logging_tracing_enabled=self.logging_tracing_enabled,
             step_dict_output_path=self.step_dict_output_path,
+            unique_id=unique_id,
         )
 
     def get_or_create_trace(self, unique_id=None):
@@ -1034,6 +1035,7 @@ class Trace(object):
         ):
             return
         trace = ''
+
 
         self.transport = (
             self.transport
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,17 +43,24 @@ def trace_transport(clean_traces):
     return epsagon.trace_factory.transport
 
 
+def init_epsagon(**kwargs):
+    default_kwargs = {
+        'token': TEST_TOKEN,
+        'app_name': TEST_APP,
+        'metadata_only': False,
+        'collector_url': TEST_COLLECTOR,
+    }
+    default_kwargs.update(kwargs)
+
+    epsagon.init(**default_kwargs)
+
+
 @pytest.fixture(scope='module', autouse=True)
-def init_epsagon():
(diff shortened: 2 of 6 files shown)


epsagon/trace.py
```diff
@@ -211,11 +211,12 @@ def _create_new_trace(self, unique_id=None):
             url_patterns_to_ignore=self.url_patterns_to_ignore,
             keys_to_ignore=self.keys_to_ignore,
             keys_to_allow=self.keys_to_allow,
-            unique_id=unique_id,
+            transport=self.transport,
```
ronyis, line 214: This is a bug fix - currently, if the user changes the default transport it won't be used in many cases

tests/conftest.py
```diff
@@ -0,0 +1,66 @@
...
+    def sent_traces(self):
+        """
+        :return: List of all the Trace objects that were sent
+        """
+        return [
+            call_args[0][0] for call_args in self.send.call_args_list
+        ]
+
+
+@pytest.fixture(scope='function', autouse=False)
```
ronyis, line 36: I wanted to use `autouse=True` but many tests in `test_trace.py` are checking the HTTP request

tests/test_trace.py
```diff
@@ -130,8 +130,15 @@ def to_dict(self):
 
 
 def setup_function(func):
+    trace_factory.use_single_trace = True
```
ronyis, line 133: Previously it used to run with multiple traces (just because the order of the tests), and some of the fixes I did here were needed to make it run like that - which is the default way

tests/test_trace.py
```diff
@@ -130,8 +130,15 @@ def to_dict(self):
...
     trace_factory.get_or_create_trace().__init__()
 
+
+@pytest.fixture(scope='function', autouse=True)
+def init_epsagon_function():
```
ronyis, line 138: Because many tests call `epsagon.init` and change the default config.

tests/test_trace.py
```diff
@@ -523,10 +503,9 @@ def test_timeout_happyflow_handler_call(wrapped_post):
 @mock.patch('requests.Session.post')
 def test_send_traces_sanity(wrapped_post):
     trace = trace_factory.get_or_create_trace()
-    trace.token = 'a'
```
ronyis, line 526: We don't support changing the token on the fly now, because the transport is created on `epsagon.init`

tests/test_trace.py
```diff
@@ -523,10 +503,9 @@ def test_timeout_happyflow_handler_call(wrapped_post):
...
     trace_factory.send_traces()
     wrapped_post.assert_called_with(
-        '',
+        'collector',
```
ronyis, line 508: Using `''` as the TC URL was not really supported

tests/test_trace.py
```diff
@@ -1328,9 +1298,9 @@ def test_send_on_error_only_with_error(wrapped_post):
 
 
 @mock.patch('requests.Session.post')
-def test_send_with_split_on_big_trace(wrapped_post):
+def test_send_with_split_on_big_trace(wrapped_post, monkeypatch):
     # Should be low enough to force trace split.
-    os.environ['EPSAGON_MAX_TRACE_SIZE'] = '500'
+    monkeypatch.setenv('EPSAGON_MAX_TRACE_SIZE', '500')
```
ronyis, line 1303: This way the env is removed also if the test fails

PR #260, opened 2020-07-29
fix(flask): fix missing response in multiple requests scenario

razlo:


tests/wrappers/test_flask_wrapper.py
```diff
@@ -85,3 +96,34 @@ def test_flask_wrapper_teardown_exception(exception_mock, _, client):
 )
 def test_lambda_wrapper_multi_thread(_):
     assert not trace_factory.use_single_trace
+
+
+def thread_a(trace_transport, client):
+    """Gets a string from A endpoint"""
+    result = client.get('/a')
+    assert trace_transport.last_trace.events[0].resource[
```
adavidai, line 104: consider extracting to a validate_response(result, 'a'), same for thread b

PR #261, opened 2020-07-30
fix(trace.py): recursively remove ignored keys

ranrib:


epsagon/trace.py
```diff
@@ -953,6 +953,15 @@ def remove_ignored_keys(self, input_dict):
...
                 if isinstance(value, dict):
                     copied_dict[key] = self.remove_ignored_keys(value)
+                else:
+                    try:
+                        json_value = json.loads(value)
+                        if isinstance(json_value, dict):
+                            copied_dict[key] = (json.dumps(
+                                self.remove_ignored_keys(json_value)
+                            ))
+                    except (TypeError, json.errors.JSONDecodeError):
```
maorlx, line 963: when can TypeError be raised?

epsagon/trace.py
```diff
@@ -953,6 +953,15 @@ def remove_ignored_keys(self, input_dict):
                 value = input_dict[key]
                 if isinstance(value, dict):
                     copied_dict[key] = self.remove_ignored_keys(value)
+                else:
+                    try:
+                        json_value = json.loads(value)
```
maorlx, line 958: Consider checking if value is str before calling json.loads

epsagon/trace.py
ranrib: Depends on the Python version you'll get either one of these

epsagon/trace.py
ranrib: Right 👍

PR #262, opened 2020-07-30
fix(trace): sampling support

LiadGoren:


epsagon/trace.py
```diff
@@ -98,6 +99,7 @@ def __init__(self):
         self.propagate_lambda_id = False
         self.logging_tracing_enabled = False
         self.step_dict_output_path = None
+        self.sample_rate = 1
```
adavidai, line 102: What's the measuring unit (1 what?) ?

epsagon/trace.py
```diff
@@ -139,6 +142,9 @@ def initialize(
             Add an epsagon log id to all loggings and prints
         :param step_dict_output_path:
             Path in the result dict to append the Epsagon steps data
+        :param sample_rate:
+            When enabled, sampling will performed performed according to the
```
adavidai, line 146:
```suggestion
            number between 0 and 1, representing the probability of each trace to be sent. When enabled (value < 1), sampling will be performed according to the
```
also, "performed performed" :)

epsagon/trace.py
```diff
@@ -483,6 +492,7 @@ def __init__(
         propagate_lambda_id=False,
         logging_tracing_enabled=False,
         step_dict_output_path=None,
+        sample_rate=1,
```
adavidai, line 495: why the default both here on Trace creation and on TraceFactory?

epsagon/utils.py
```diff
@@ -98,6 +98,7 @@ def init(
     propagate_lambda_id=False,
     logging_tracing_enabled=True,
     step_dict_output_path=None,
+    sample_rate=1,
```
adavidai, line 101: yet another default :) 
Please create a const DEFAULT_SAMPLE_RATE and use it instead of hard code it in different places.

epsagon/utils.py
```diff
@@ -122,9 +123,10 @@ def init(
     :param logging_tracing_enabled: Add an epsagon log id to logging calls
     :param step_dict_output_path:
         Path in the result dict to append the Epsagon steps data
+    :param sample_rate:
+        When enabled, sampling will be performed according to the given value.
```
adavidai, line 127:
```suggestion
       number between 0 and 1, representing the probability of each trace to be sent. When enabled (value < 1), sampling will be performed according to the given value.
```

tests/test_epsagon_init.py
```diff
@@ -144,6 +144,7 @@ def test_epsagon_wrapper_env_init(wrapped_get):
     'EPSAGON_LOGGING_TRACING_ENABLED': 'TRUE',
     'AWS_LAMBDA_FUNCTION_NAME': None,
     'EPSAGON_STEPS_OUTPUT_PATH': '',
+    'EPSAGON_SAMPLE_RATE': 0.5
```
adavidai, line 147: Extract 0.5 here and in the test assertion to a const (making it easier to understand the connection between the two)

adavidai:
BTW, I think this is a "feat(trace): ..." (and not a fix)

epsagon/trace.py
LiadGoren: 1. as you can see - all parameters are being set to defaults both in Trace and TraceFactory.
It's probably not ideal, as well as it's not ideal to hand over more than 20 parameters to a constructor, but it is how it is for now.
2. There is at least one more place that initialises Trace, besides TraceFactory. Won't the the chance to break anything.
It's just a default value.

tests/test_epsagon_init.py
LiadGoren: These tests are pretty straightforward... values in the mock object should match the values called in the init function...
All other parameters are not stored as consts as well (and I also think it should stay like that since these are tests)

LiadGoren:
@adavidai Please provide a use-example for the new fixture in tests

adavidai:
@LiadGoren sure, there's one in @razlo 's open PR

epsagon/trace.py
adavidai: Opened a ticket to look into refactoring that

PR #266, opened 2020-08-05
tests: add django wrapper testing

razlo:


tests/wrappers/test_django_wrapper.py
```diff
@@ -0,0 +1,78 @@
...
+    right content.
+    """
+    request_mw = DjangoRequestMiddleware(test_request)
+
+    request_mw.before_request()
+    request_mw.after_request({"content": "bla"})
+
+    runner_mock.assert_called_once()
+    trigger_mock.factory.assert_called_once()
+    assert trace_transport.last_trace.events[
```
maorlx, line 61: validate method & path equals the ones defined in the test_request fixture

tests/wrappers/test_django_wrapper.py
```diff
@@ -0,0 +1,78 @@
...
+        0].resource['metadata']['Request Data'] == 'test_body'
```
maorlx, line 62: I would put the test_body into a const, as its being used here twice

tests/wrappers/test_django_wrapper.py
razlo: we currently don't capture the method :(
added validation for the body

PR #268, opened 2020-08-06
feat(trace.py): switch to urllib3

ranrib:


epsagon/utils.py
```diff
@@ -260,7 +260,11 @@ def collect_container_metadata(metadata):
     metadata_uri = os.environ.get('ECS_CONTAINER_METADATA_URI')
     if not metadata_uri:
         return
-    container_metadata = json.loads(requests.get(metadata_uri).content)
+
+    session = urllib3.PoolManager()
```
ronyis, line 264: Are you sure it's good to create a `PoolManager` object each time?

PR #273, opened 2020-08-27
feat(sqlalchemy.py): add support for sqlalchemy

ranrib:
Screen Shot 2020-08-27 at 16 02 12


epsagon/utils.py
```diff
@@ -29,6 +29,20 @@
...
 
+def patch_once(patch_module, patch_name, wrapper):
+    """
+    Patches a module once by setting `EPSAGON_MARKER` on the wrapped object
+    :param patch_module: module to patch
+    :param patch_name: attribute name to path
+    :param wrapper: new wrapper to set
+    :return: None
+    """
+    (_, _, original) = wrapt.resolve_path(patch_module, patch_name)
```
maorlx, line 40: no need for the ( )

epsagon/events/sqlalchemy.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # override event type with the specific DB type
+        self.resource['type'] = database_connection_type(
+            instance.bind.url.host,
+            self.RESOURCE_TYPE
+        )
+
+        if exception is not None:
+            self.set_exception(exception, traceback.format_exc())
+
```
maorlx, line 53: don't we wanna save the response? (if metadata only is false)

ranrib: There's nothing interesting in the response here

epsagon/utils.py
ranrib: Thanks, fixed!

epsagon/events/sqlalchemy.py
maorlx: right, we just catch init & close sessions

maorlx (approved):

PR #278, opened 2020-08-31
feat(custom.py): add @epsagon.measure

ranrib:


tests/wrappers/test_custom.py
```diff
@@ -0,0 +1,24 @@
...
+        return retval
+
+    @epsagon.python_wrapper(name='test-func')
+    def wrapped_function():
+        measured_function()
+        return retval
+
+    assert wrapped_function() == retval
+    labels = json.loads(trace_transport.last_trace.events[0].resource['metadata']['labels'])
+    assert labels['measured_function_duration'] == 0
```
maorlx, line 24: I would mocked function so the tested measure time will be > 0

ranrib: Done 👍

maorlx (dismissed):

maorlx (approved):

PR #280, opened 2020-09-05
feat(trace.py) get trace url

ranrib:


adavidai (dismissed):

epsagon/trace.py
```diff
@@ -1162,6 +1161,26 @@ def _send_traces(self):
...
                 print('trace:', trace)
 
+    def get_trace_url(self):
+        """
+        Return the trace URL based on the runner ID.
+        Handling special case for Lambda functions.
+        :return: string of the trace URL.
+        """
+        trace_url = TRACE_URL_PREFIX.format(
+            id=self.runner.resource['metadata'].get('trace_id', ''),
```
adavidai, line 1171: Maybe verify there's a runner, and it has metadata? (So this will not crash in "edge cases")

adavidai (dismissed):
See a minor comment below

epsagon/trace.py
ranrib: Fixed and added a test 👍

adavidai (approved):

PR #281, opened 2020-09-08
feat(aiohttp.py): support aiohttp

ranrib:


epsagon/runners/aiohttp.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    def update_response(self, response):
+        """
+        Adds response data to event.
+        :param response: WSGI Response
+        :return: None
+        """
+        body = response.body
+        if  response.body.__class__.__name__ == 'StringPayload':
```
maorlx, line 66: use isinstance instead

epsagon/runners/aiohttp.py
```diff
@@ -0,0 +1,84 @@
...
+            dict(request.headers)
+        )
+
+    def update_response(self, response):
+        """
+        Adds response data to event.
+        :param response: WSGI Response
+        :return: None
+        """
+        body = response.body
```
maorlx, line 65: do that only if the next condition is false

epsagon/runners/aiohttp.py
```diff
@@ -0,0 +1,84 @@
...
+        add_data_if_needed(
+            self.resource['metadata'],
+            'Response Data',
+            body.decode('utf-8')
+        )
+
+        add_data_if_needed(
+            self.resource['metadata'],
+            'Response Headers',
+            dict(response.headers)
```
maorlx, line 78: reponse.headers can be None? if so then should check it

epsagon/wrappers/aiohttp.py
```diff
@@ -0,0 +1,61 @@
...
+    if ignore_request('', request.path.lower()):
+        return await handler(request)
+
+    trace = epsagon.trace.trace_factory.get_or_create_trace()
+    trace.prepare()
+    runner = None
+    response = None
+
+    try:
+        body = await request.text()
```
maorlx, line 39: 1. When happens if another client is initialized during this await? or await handler(request)?
I think it might cause another tracer to be initialized, so the current tracer will be overriden. Maybe using the trace object itself will solve that issue, not counting on the trace factory (multile_traces mode is not good here as its all running on the same thread)
2. Should this feature support aiohttp server? if so, we might do want multiple traces support

epsagon/runners/aiohttp.py
ranrib: I would need to import it from aiohttp, which doesn't exist in any environment, which means that this will be in a try-catch block - which looks much worse IMO. (5-6 lines instead of one)
If you still prefer it, let me know and I'll do that

epsagon/runners/aiohttp.py
ranrib:
```
body = response.body
if  response.body.__class__.__name__ == 'StringPayload':
        body = getattr(response.body, '_value')
```
Is just the same as
```
if  response.body.__class__.__name__ == 'StringPayload':
        body = getattr(response.body, '_value')
else:
        body = response.body
```

epsagon/wrappers/aiohttp.py
ranrib: The middleware is already co-routined async-wise, so it doesn't reach the same tracer together. That's also the reason to wrap the @middleware decorator

epsagon/runners/aiohttp.py
maorlx: I meant by saving the body=response.body assignment

epsagon/wrappers/aiohttp.py
maorlx: not sure I get why - the trace factory is using a single tracer (default mode), which the middleware doesn't know about
In flask & django, this state ("multiple traces" by thread ID) is enabled by our instrumentation automatically

ranrib: You're right, I switched to multiple traces mode

epsagon/runners/aiohttp.py
ranrib: Fixed

epsagon/wrappers/aiohttp.py
ranrib: Fixed

maorlx (approved):

maorlx (approved):

PR #284, opened 2020-09-21
fix(logging.py): fix add trace id to logs flags

razlo:
Fixing the flag set for trace factory.
What happened is that logging.py patch was called on import epsagon, so when we checked if logging_enabled we always got FALSE - no matter what was the given flag value.


ranrib:
@razlo @adavidai

PR #299, opened 2020-11-26
feat(botocore): support eventbridge

idonava:


epsagon/events/botocore.py
```diff
@@ -117,13 +117,82 @@ def update_response(self, response):
...
+        :param start_time: Start timestamp (epoch)
+        :param response: response data
+        :param exception: Exception (if happened)
+        """
+
+        super(BotocoreCloudWatchEvent, self).__init__(
+            wrapped, instance, args, kwargs, start_time, response, exception
+        )
+        _, request_data = args
+        entries = request_data.get('Entries', [{}])[0]
```
maorlx, line 155: I would change it so you initially get 'Entries' and then check it (so it won't explode if request_data.get('Entries') is empty)

epsagon/events/botocore.py
```diff
@@ -117,13 +117,82 @@ def update_response(self, response):
...
+        self.resource['name'] = (
```
maorlx, line 156: can do entries.get('EventBusName', 'Cloudwatch Events')

epsagon/events/botocore.py
```diff
@@ -117,13 +117,82 @@ def update_response(self, response):
...
             response['ResponseMetadata']['HTTPStatusCode']
 
 
+class BotocoreCloudWatchEvent(BotocoreEvent):
+    """
+    Represents cloudwatch events (eventbridge) botocore event.
+    """
+
+    RESOURCE_TYPE = 'eventbridge'
+    RESOURCE_TYPE_UPDATE = 'events'
```
maorlx, line 135: not important but would rename to event (non plural)

epsagon/events/botocore.py
```diff
@@ -117,13 +117,82 @@ def update_response(self, response):
...
+    def update_response(self, response):
+        """
+        Adds response data to event.
+        :param response: Response from botocore
+        :return: None
+        """
+        super(BotocoreCloudWatchEvent, self).update_response(response)
+        if self.resource['operation'] == 'PutEvents' and 'Entries' in response:
+            self.resource['metadata']['aws.cloudwatch.event_id'] = \
+                response['Entries'][0]['EventId']
```
maorlx, line 194: it would be good to make sure response['Entries'] is not empty

epsagon/triggers/aws_lambda.py
```diff
@@ -439,7 +439,10 @@ def __init__(self, start_time, event, context):
         super(EventsLambdaTrigger, self).__init__(start_time)
 
         self.event_id = str(event['id'])
-        self.resource['name'] = str(event['resources'][0].split('/')[-1])
+        name = 'CloudWatch Events'
```
maorlx, line 442: in case of events bridge event, what would be the resource name for the event?

maorlx (changes requested):

epsagon/events/botocore.py
idonava: 'events' is the aws resource name `boto3.client('events')...`

epsagon/triggers/aws_lambda.py
idonava: the trigger resource name will be always 'CloudWatch Events' because EventBusName doesn't exist here.

maorlx (approved):

PR #302, opened 2020-12-06
Fastapi ep 5296

maorlx:
Docs will be updated after this PR will be merged

### DIFF
--- a/epsagon/modules/fastapi.py
+++ b/epsagon/modules/fastapi.py
@@ -22,6 +22,7 @@ def _wrapper(wrapped, _instance, args, kwargs):
         return wrapped(*args, **kwargs)
     route_class = kwargs.get('route_class', APIRoute)
     if route_class != APIRoute:
+        # custom routes are not supported
         return wrapped(*args, **kwargs)
     kwargs['route_class'] = TracingAPIRoute
     return wrapped(*args, **kwargs)
--- a/epsagon/modules/fastapi.py
+++ b/epsagon/modules/fastapi.py
@@ -23,7 +23,9 @@ def _wrapper(wrapped, _instance, args, kwargs):
     route_class = kwargs.get('route_class', APIRoute)
     if route_class != APIRoute:
         # custom routes are not supported
-        print_debug('Custom FastAPI routes are not supported')
+        print_debug(
+            f'Custom FastAPI route {route_class.__name__} is not supported'
+        )
         return wrapped(*args, **kwargs)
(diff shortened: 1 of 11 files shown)


epsagon/modules/fastapi.py
```diff
@@ -0,0 +1,40 @@
...
+    :param args: wrapt's args
+    :param kwargs: wrapt's kwargs
+    """
+
+    # Skip on Lambda environment since it's not relevant and might be duplicate
+    if is_lambda_env():
+        return wrapped(*args, **kwargs)
+    route_class = kwargs.get('route_class', APIRoute)
+    if route_class != APIRoute:
+        # custom routes are not supported
```
adavidai, line 25: How about a debug print in that case - "custom fastapi route ignored", along with whatever data we have on the route?

epsagon/wrappers/fastapi.py
```diff
@@ -0,0 +1,77 @@
...
+            trace.prepare()
+            runner = None
+            response = None
+            try:
+                body = await request.json()
+            except json.decoder.JSONDecodeError:
+                body = ''
+            try:
+                runner = FastapiRunner(time.time(), request, json.dumps(body))
+                trace.set_runner(runner)
```
adavidai, line 52: Is it possible that more than one instance does set_runner on the same active trace, overriding each other?

adavidai (commented):
Looks good 👍
See some comments inline

epsagon/wrappers/fastapi.py
maorlx: no, since the callbacks will be called in different jobs and our trace module will know to create a different trace for each job 
See: `epsagon.trace.trace_factory.switch_to_async_tracer()`

tests/wrappers/test_fastapi_wrapper.py
```diff
@@ -77,3 +87,30 @@ async def test_fastapi_exception(_, trace_transport, fastapi_app):
     assert runner.error_code == ErrorCode.EXCEPTION
     assert runner.exception['type'] == 'Exception'
     assert runner.exception['message'] == 'test'
+
+
+@pytest.mark.asyncio
+#@asynctest.patch('epsagon.trace.trace_factory.use_async_tracer')
```
adavidai, line 93: remove?

adavidai (approved):

adavidai (dismissed):

adavidai (approved):

PR #307, opened 2020-12-17
fix(trace.py): ut breaks in async mode

henprtz:


adavidai (approved):

adavidai:
@henprtz PR name should follow semantic release or a new release will not be published.

PR #308, opened 2020-12-20
Fixing fast api errors handling

maorlx:


henprtz (changes requested):
Please also wrap the entire logic of lines 60-75 to capture the library exception but maintain regular flow for the user code

henprtz (approved):

epsagon/wrappers/fastapi.py
```diff
@@ -53,24 +61,31 @@ async def custom_route_handler(request: Request) -> Response:
...
+                            ''
+                    ):
+                        return response
+
+                    runner.update_response(response)
 
-                runner.update_response(response)
+                if runner:
+                    epsagon.trace.trace_factory.send_traces()
+            except Exception as exception:  # pylint: disable=W0703
```
henprtz, line 84: Let's print something here so we'll be aware if such issue occurs

maorlx: done

epsagon/wrappers/fastapi.py
```diff
@@ -53,24 +61,31 @@ async def custom_route_handler(request: Request) -> Response:
...
+                            response.headers.get('Content-Type', '').lower(),
+                            ''
+                    ):
+                        return response
 
-                runner.update_response(response)
+                    runner.update_response(response)
+
+                if runner:
+                    epsagon.trace.trace_factory.send_traces()
```
adavidai, line 83: Where is `send_traces()` done in case that there was a `raised_err`?
If it is not needed maybe we can just raise raised_err in the except block (line #71), instead of adding it to the if and then checking for it and raising later?

maorlx: the send traces will be done even if raised_err :) only the response update won't be performed if raised_err

adavidai: right, my bad :)

adavidai (approved):

PR #314, opened 2021-01-10
fix(tornado.py): fix tornado mis-handle of 404s

ranrib:
In addition:
1. capturing redis response
2. support older redis versions
3. fixing edge case in urllib3
4. fixing log-correlation that was always handled as False (since trace_factory was not initialized)
5. adding some debug prints to Tornado

### DIFF
--- a/epsagon/modules/tornado.py
+++ b/epsagon/modules/tornado.py
@@ -92,7 +92,8 @@ class TornadoWrapper(object):
             tornado_runner = cls.RUNNERS.pop(unique_id)
 
             # Ignoring 404s
-            if getattr(instance, '._status_code', None) == 404:
+            if getattr(instance, '_status_code', None) == 404:
+                print_debug('Ignoring 404 Tornado request')
                 return res
 
             trace = epsagon.trace.trace_factory.switch_active_trace(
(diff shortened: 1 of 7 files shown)


epsagon/modules/tornado.py
```diff
@@ -80,6 +91,10 @@ def after_request(cls, wrapped, instance, args, kwargs):
 
             tornado_runner = cls.RUNNERS.pop(unique_id)
 
+            # Ignoring 404s
+            if getattr(instance, '._status_code', None) == 404:
```
Alon-Katz, line 95: Is the dot in the attribute name is correct? shouldn't it be just '_status_code'

epsagon/utils.py
```diff
@@ -95,6 +95,24 @@ def get_tc_url(use_ssl):
...
 
+def get_trace_log_config():
+    """Returns the trace log correlation configuration"""
+    # Default is True
+    logging_tracing_enabled = True
+
+    # If EPSAGON_LOGGING_TRACING_ENABLED exists as an env var - use it
+    if os.getenv('EPSAGON_LOGGING_TRACING_ENABLED'):
+        logging_tracing_enabled = (
+            os.getenv('EPSAGON_LOGGING_TRACING_ENABLED') or ''
```
Alon-Katz, line 106: I think the `or ''` isn't required because you already checked for a value in EPSAGON_LOGGING_TRACING_ENABLED inside the if statement

Alon-Katz (approved):

Alon-Katz (approved):

PR #315, opened 2021-01-10
fix(tornado.py): fix response parsing

ranrib:


epsagon/runners/tornado.py
```diff
@@ -96,6 +96,10 @@ def update_response(self, response, response_body=None):
         )
 
         if response_body:
+            if isinstance(response_body, list):
+                response_body = response_body[0]
```
adavidai, line 100: Is it possible that the passed response_body is the actual response that the user uses? 
Could be a good idea to use a different var and not **change** the response_body itself

epsagon/runners/tornado.py
```diff
@@ -96,6 +96,10 @@ def update_response(self, response, response_body=None):
...
+            if isinstance(response_body, bytes):
+                response_body = response_body.decode('utf-8')
```
adavidai, line 102: same

epsagon/runners/tornado.py
adavidai: Also handle empty list use case

ranrib:
Addressed in https://github.com/epsagon/epsagon-python/pull/317

PR #317, opened 2021-01-12
fix(tornado.py): fix mishandling of 404s

ranrib:
In addition:

1. capturing redis response
2. support older redis versions
3. Fixing mishandling of 404 in aiohttp
4. fixing log-correlation that was always handled as False (since trace_factory was not initialized)
5. adding some debug prints to Tornado

### DIFF
--- a/epsagon/modules/redis.py
+++ b/epsagon/modules/redis.py
@@ -44,6 +44,12 @@ def patch():
         'Redis.execute_command',
         _single_wrapper
     )
+    # Version < 3.0
+    wrapt.wrap_function_wrapper(
+        'redis',
+        'StrictRedis.execute_command',
+        _single_wrapper
+    )
     wrapt.wrap_function_wrapper(
         'redis.client',
         'Pipeline.immediate_execute_command',
--- a/epsagon/modules/tornado.py
+++ b/epsagon/modules/tornado.py
@@ -92,7 +92,8 @@ class TornadoWrapper(object):
             tornado_runner = cls.RUNNERS.pop(unique_id)
 
             # Ignoring 404s
-            if getattr(instance, '._status_code', None) == 404:
+            if getattr(instance, '_status_code', None) == 404:
+                print_debug('Ignoring 404 Tornado request')
                 return res
 
             trace = epsagon.trace.trace_factory.switch_active_trace(
--- a/epsagon/modules/tornado.py
+++ b/epsagon/modules/tornado.py
@@ -11,7 +11,7 @@ import wrapt
 import epsagon.trace
 from epsagon.runners.tornado import TornadoRunner
 from epsagon.http_filters import ignore_request, is_ignored_endpoint
-from epsagon.utils import collect_container_metadata
+from epsagon.utils import collect_container_metadata, print_debug
 
 TORNADO_TRACE_ID = 'epsagon_tornado_trace_key'
 
@@ -31,6 +31,7 @@ class TornadoWrapper(object):
         :param args: wrapt's args
         :param kwargs: wrapt's kwargs
         """
+        print_debug('before_request Tornado request')
         try:
             ignored = ignore_request('', instance.request.path)
             if not ignored and not is_ignored_endpoint(instance.request.path):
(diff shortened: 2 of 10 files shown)


epsagon/utils.py
```diff
@@ -95,6 +95,19 @@ def get_tc_url(use_ssl):
...
+    """Returns the trace log correlation configuration"""
+    # In case we're running on AWS Lambda, logging correlation is disabled
+    if is_lambda_env():
+        return False
+
+    # If EPSAGON_LOGGING_TRACING_ENABLED exists as an env var - use it
+    if os.getenv('EPSAGON_LOGGING_TRACING_ENABLED'):
+        return os.getenv('EPSAGON_LOGGING_TRACING_ENABLED').upper() == 'TRUE'
+
+    return True
```
maorlx, line 108:
```suggestion
    return os.getenv('EPSAGON_LOGGING_TRACING_ENABLED', 'TRUE').upper() == 'TRUE'
```

epsagon/wrappers/aiohttp.py
```diff
@@ -38,16 +41,24 @@ async def AiohttpMiddleware(request, handler):
...
     except Exception as exception: # pylint: disable=W0703
         warnings.warn('Could not extract request', EpsagonWarning)
 
     raised_err = None
     try:
         response = await handler(request)
+        print_debug('[aiohttp] got response')
     except Exception as exception:  # pylint: disable=W0703
+        # Ignoring 404s
+        if type(exception).__name__ == 'HTTPNotFound':
```
maorlx, line 58: can we import the specific exception class and compare the types objects?

epsagon/modules/redis.py
```diff
@@ -44,6 +44,12 @@ def patch():
         'Redis.execute_command',
         _single_wrapper
     )
+    # Version < 3.0
+    wrapt.wrap_function_wrapper(
```
maorlx, line 48: to make sure, in newer versions this call won't explode - right?

epsagon/modules/tornado.py
```diff
@@ -80,6 +91,11 @@ def after_request(cls, wrapped, instance, args, kwargs):
 
             tornado_runner = cls.RUNNERS.pop(unique_id)
 
+            # Ignoring 404s
+            if getattr(instance, '_status_code', None) == 404:
+                print_debug('Ignoring 404 Tornado request')
```
maorlx, line 96: should pop the trace from trace_factory (and prevent memory leaks for those 404 traces)

epsagon/runners/tornado.py
```diff
@@ -69,6 +69,19 @@ def __init__(self, start_time, request):
...
+                body = request.body
+                if isinstance(body, bytes):
+                    body = body.decode('utf-8')
+                add_data_if_needed(
+                    self.resource['metadata'],
+                    'Request Data',
+                    body
+                )
+        except Exception as exception:  # pylint: disable=broad-except
+            print_debug('Could not extract body: {}'.format(exception))
```
maorlx, line 83:
```suggestion
            print_debug('Could not extract request body: {}'.format(exception))
```

epsagon/runners/tornado.py
```diff
@@ -83,10 +96,16 @@ def update_response(self, response, response_body=None):
         )
 
         if response_body:
+            body = response_body
+            if isinstance(body, list):
+                body = body[0]
```
maorlx, line 101: in which cases is the body list? (wondering whether we want to collect all the list items)

maorlx (changes requested):

epsagon/runners/tornado.py
ranrib: It seems to be always a list from what I've seen, and always a single element

epsagon/modules/redis.py
ranrib: That's right, this code is failsafe

epsagon/utils.py
ranrib: <3

epsagon/wrappers/aiohttp.py
ranrib: <3

maorlx (dismissed):

maorlx (approved):

PR #322, opened 2021-01-25
feat(dbapi.py): obfuscate sql queries

ronnathaniel:
Original:
INSERT   INTO people   VALUES ('Mahatma', 'Ghandi', 78);
SELECT Count(*)   FROM people   WHERE first='M' OR first NOT IN       (SELECT last FROM people WHERE age =30 AND last='F')   AND NOT EXISTS       (SELECT age FROM people WHERE first='J');"

Obfuscated:
INSERT   INTO people   VALUES (???, ???, ???);
SELECT Count(*)   FROM people   WHERE first=??? OR first NOT IN       (SELECT last FROM people WHERE age=??? AND last=???)   AND NOT EXISTS       (SELECT age FROM people WHERE first=???);


epsagon/utils.py
```diff
@@ -445,3 +453,74 @@ def database_connection_type(hostname, default_type):
     if 'redshift.amazonaws' in hostname:
         return 'redshift'
     return default_type
+
+
```
maorlx, line 457: Please add tests to this function (see tests/test_utils)

epsagon/utils.py
```diff
@@ -445,3 +453,74 @@ def database_connection_type(hostname, default_type):
...
+        for i, value in enumerate(values[:]):
+            parts = list(value.partition(signal))
+            # if could not match signal correctly, leave it
+            if not parts[1]:
+                continue
+            parts[0] = parts[0].strip()
+            parts[2] = replacer
+            values[i] = ''.join(parts)
+    else:
+        values = [replacer for _ in values[:]]
```
maorlx, line 520: you can remove the [:], it will assign the new list after this statement

epsagon/utils.py
```diff
@@ -445,3 +453,74 @@ def database_connection_type(hostname, default_type):
...
+    Obfuscate SQL queries to protect sensitive uploads
+    :param query: The SQL query string
+    :param operation: Operation in DB
+    :return: the obfuscated query (string)
+    :var bounds: specifier for locating values based upon operation
+    :var bounds: start is leftmost bound, stop is rightmost bound
+    :var bounds: separator is the symbol between values
+    :var bounds: signal is a preceding identifier for each value
+    """
+    bounds = {
```
maorlx, line 469: this parsing indeed works for some statements, but will fail on more complex scenarios - for example "select * from bla where x=1 ORDER..."  or for example "select * from bla where EXISTS (.... another statement..)
You can either: 
- catch those cases so you won't fail on obfuscating the string (I don't recommend that as it can get very complicated)
- I would suggest to look at sqlparse - it seems it will help parsing the sql query: (I didn't take a deep look though)
https://sqlparse.readthedocs.io/en/latest/analyzing/

maorlx (commented):
Thanks for adding the feature! I left some comments inline

epsagon/events/dbapi.py
```diff
@@ -120,7 +120,12 @@ def __init__(
                 (operation == 'select') or
                 (not trace_factory.metadata_only)
         ):
-            self.resource['metadata']['Query'] = query[:MAX_QUERY_SIZE]
+
+            query_max = query[:MAX_QUERY_SIZE]
```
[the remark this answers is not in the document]
ronnathaniel: Because we want to conditionally obfus the maxed query, and not the other way around.
I think it should only touch self.resource[meta][query] after the entire query is prepared

epsagon/utils.py
```diff
@@ -231,17 +234,21 @@ def init(
...
+            | split_on_send
         ),
         propagate_lambda_id=(
-                ((os.getenv('EPSAGON_PROPAGATE_LAMBDA_ID') or '').upper() ==
-                 'TRUE')
-                | propagate_lambda_id
+            ((os.getenv('EPSAGON_PROPAGATE_LAMBDA_ID') or '').upper() ==
+             'TRUE')
+            | propagate_lambda_id
+        ),
```
ronnathaniel, line 244: all this pycharm did automatically. if it passes the tests it it good enough or should be reverted?

epsagon/utils.py
ronnathaniel: I looked into every sql parsing library, none of them are good enough for what we want (which is still pretty simple).
I think i'll go the first route, and only obfuscate when 100% safe

epsagon/utils.py
ronnathaniel: added tests

epsagon/utils.py
ronnathaniel: I ended up using a bit of sqlparse & also ignoring cases we don't want to obfuscate. Looks much stronger now, and can handle very complex clauses.

epsagon/events/dbapi.py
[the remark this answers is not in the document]
maorlx: I meant keeping the same logic but just using the same variable name as query_max is a redundant var (anyway, its not important)

epsagon/events/dbapi.py
```diff
@@ -108,7 +108,12 @@ def __init__(
...
+            if trace_factory.obfuscate_sql:
+                query_max = obfuscate_sql_query(query_max, operation)
```
maorlx, line 114: lets wrap this call with try except and in case of failures use print_debug with a relevant message

requirements-dev.txt
```diff
@@ -11,6 +11,7 @@ pytest==6.0.0; python_version >= '3.5'
 pytest==4.6.0; python_version < '3.5'
 requests
 sqlalchemy
+sqlparse
```
maorlx, line 14: needed also at requirements

epsagon/utils.py
```diff
@@ -445,3 +456,123 @@ def database_connection_type(hostname, default_type):
...
+                values[i] = v
+
+            values = ''.join(values)
+
+            t = t[:positions[0]] \
+                + values \
+                + t[positions[1]:]
+
+        obfuscated_query.append(t)
+
```
maorlx, line 575: try to split this function as it became too long :) (the internal logic for example can be moved)

epsagon/utils.py
```diff
@@ -445,3 +456,123 @@ def database_connection_type(hostname, default_type):
...
+    # order not preserved in py2. Dont default to OrderedDict,not inefficient
+    to_dict = collections.OrderedDict if sys.version_info.major == 2 else dict
+
+    [token_type,
+    start, stop,
+    separators, operators] = to_dict(bounds[operation]).values()
+    obfuscated_query = []
+
+    parse_query = sqlparse.parse(query)[0].tokens
```
maorlx, line 525: it might be possible the query will include more than one statement (returned from sqlparse)
In that case, make sure to skip the query (and print_debug about it)

epsagon/utils.py
```diff
@@ -445,3 +456,123 @@ def database_connection_type(hostname, default_type):
...
+        query = query.decode('UTF-8')
+
+    if isinstance(operation, bytes):
+        operation = operation.decode('UTF-8')
+
+    if operation not in bounds.keys():
+        return query
+
+    # order not preserved in py2. Dont default to OrderedDict,not inefficient
+    to_dict = collections.OrderedDict if sys.version_info.major == 2 else dict
```
maorlx, line 518: it would be better to assign the values manually without that

maorlx (changes requested):

epsagon/events/dbapi.py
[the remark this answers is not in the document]
ronnathaniel: I understand now. Definitely agree, will update

epsagon/utils.py
```diff
@@ -455,29 +456,49 @@ def database_connection_type(hostname, default_type):
...
+    :return: string type, compatible for multiple options of regex split
+    """
+    if isinstance(options, list):
+        options = '|'.join(map(re.escape, options))
+
+    # if options is not a string by now param was invalid
+    if not isinstance(options, str):
+        return ''
+
+    return f'({options})'
```
maorlx: supported only in python3

PR #339, opened 2021-04-26
fix(aws lambda wrapper): allow exception without type

adavidai:


epsagon/event.py
```diff
@@ -114,7 +114,10 @@ def set_exception(
         :param from_logs: True if the exception was captured from logging
         """
         self.error_code = ErrorCode.EXCEPTION
-        self.exception['type'] = type(exception).__name__
+        if hasattr(exception, '__name__'):
```
ronyis, line 117:
```suggestion
        if hasattr(type(exception), '__name__'):
```

epsagon/event.py
```diff
@@ -114,7 +114,10 @@ def set_exception(
...
+            self.exception['type'] = type(exception).__name__
+        else:
+            self.exception['type'] = None
```
ronyis, line 120:
```suggestion
            self.exception['type'] = ''
```

maorlx (approved):

PR #340, opened 2021-04-28
fix(flask.py): ignore endpoints correctly

ronnathaniel:


epsagon/wrappers/flask.py
```diff
@@ -34,9 +34,11 @@ def __init__(self, app, ignored_endpoints=None):
             return
         setattr(app, self.EPSAGON_MARKER, True)
         self.app = app
-        self.ignored_endpoints = []
         if ignored_endpoints:
-            self.ignored_endpoints = ignored_endpoints
+            if isinstance(ignored_endpoints, list):
```
ranrib, line 38: We can drop this - no one uses this param

ronnathaniel: just in case its passed in later on - itll be sure to include it to the std ignored endpoints

ranrib: I would prefer to not maintain 2 scenarios just for Flask. It can just lead to confusion / errors. It's better to use the standard way of setting ignored endpoints

maorlx (approved):

PR #342, opened 2021-05-05
Feat (Cognito) add support for ops: AdminInitiateAuth, AdminRespondToAuthChallenge, SignUp 

ronnathaniel:
Screen Shot 2021-05-05 at 6 07 36 PM
Screen Shot 2021-05-05 at 6 07 46 PM
Screen Shot 2021-05-05 at 6 08 07 PM

### DIFF
--- a/epsagon/events/botocore.py
+++ b/epsagon/events/botocore.py
@@ -1090,11 +1090,13 @@ class BotocoreCognitoEvent(BotocoreEvent):
 
         self.OPERATION_TO_FUNC.update({
             'AdminCreateUser': self.admin_create_user_op,
+            'AdminInitiateAuth': self.admin_initiate_auth_op,
             'AdminListGroupsForUser': self.admin_list_user_group_op,
             'AdminSetUserPassword': self.admin_set_pass_op,
             'DescribeUserPool': self.describe_user_pool_op,
             'ListUsers': self.list_users_op,
             'UpdateUserPool': self.update_pool_op,
+            'SignUp': self.sign_up_op,
         })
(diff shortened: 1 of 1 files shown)


ronnathaniel:
Not to worry. Lint will be fixed

epsagon/events/botocore.py
```diff
@@ -1221,6 +1226,39 @@ def update_pool_op(self, args, _):
...
+    def admin_respond_to_auth_challenge_op(self, args, _):
+        """
+        Process Respond to Auth Challenge as Admin
+        :param args: command arguments
+        :param _: unused, kwargs
+        :return: None
+        """
+        _, request_args = args
+        self.resource['name'] = request_args['UserPoolId']
+        self.resource['metadata'].update(request_args)
```
ophiryael, line 1260: Same exact function logic (admin_initiate_auth_op &  admin_respond_to_auth_challenge_op) and both are duplicates of update_pool_op.

epsagon/events/botocore.py
```diff
@@ -2,7 +2,7 @@
 botocore events module.
 """
 
-# pylint: disable=C0302
+# pylint: disable=C0301, C0302
```
ophiryael, line 5: Not sure if it's needed at all, but if it is, fix the line length instead of ignoring the linter (unless there is an actual issue).

ophiryael:
I think the CR fixes changed functionality of existing operations. Let's talk about it on Slack.

ophiryael (approved):
Looks good

PR #347, opened 2021-05-23
Adding custom api routes support

maorlx:


epsagon/wrappers/fastapi.py
```diff
@@ -248,21 +247,27 @@ def wrapped_handler(*args, **kwargs):
...
+    :param kwargs: wrapt's kwargs
     """
+    result = wrapped(*args, **kwargs)
+    # Skip on Lambda environment since it's not relevant and might be duplicate
+    if not is_lambda_env() and instance:
+        try:
+            if instance.dependant and instance.dependant.call:
+                _wrap_handler(
+                    instance.dependant,
+                    kwargs.get('status_code', DEFAULT_SUCCESS_STATUS_CODE)
```
adavidai, line 265: Maybe the default status code should be UNKNOWN or something that will differentiate it from cases where we could get the status code and it was successful?

adavidai (approved):

epsagon/wrappers/fastapi.py
maorlx: its actually our default value to match the FastAPI default endpoint status code

PR #356, opened 2021-06-01
feat(urrlib3): collecting payload

henprtz:


tests/test_utils.py
```diff
@@ -14,6 +14,7 @@ def test_blacklist_url():
     :return: None
     """
 
+    original = epsagon.http_filters.BLACKLIST_URLS # Storing old state
```
adavidai, line 17: Extract to pytest.fixture like: 
```
original = epsagon.http_filters.BLACKLIST_URLS
yield
epsagon.http_filters.BLACKLIST_URLS = original
```
(instead of making it a part of this specific test)

tests/modules/test_urllib3.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_data_capture_on_preload_content_true(trace_transport):
+    @epsagon.wrappers.python_function.python_wrapper
+    def wrapped_function():
+        headers = {
+            'Content-Type': 'application/json'
+        }
+
+        http = urllib3.PoolManager()
+        return http.request('POST', TEST_URL, headers=headers, body='', preload_content=True)
```
adavidai, line 49: 1. `POST`? (was expecting a GET)
2. What happens if preload_content=False?

tests/modules/test_urllib3.py
```diff
@@ -0,0 +1,59 @@
...
+    @epsagon.wrappers.python_function.python_wrapper
+    def wrapped_function():
+        headers = {
+            'Content-Type': 'application/json'
+        }
+
+        http = urllib3.PoolManager()
+        conn = http.connection_from_url(TEST_URL)
+        urllib_response = conn.urlopen(
+            method='POST',
```
adavidai, line 24: `GET`?

tests/modules/test_urllib3.py
```diff
@@ -0,0 +1,59 @@
...
+        headers = {
+            'Content-Type': 'application/json'
+        }
+
+        http = urllib3.PoolManager()
+        return http.request('POST', TEST_URL, headers=headers, body='', preload_content=True)
+
+    response = wrapped_function()
+    data = response.read()
+    # In this case data will not be available in the buffer
```
adavidai, line 53: Is this a valid behavior (does it behave the same way without the Epsagon wrapper)? 
I suggest a different test - can we run the same request twice, once with Epsagon wrapper and once without it, and compare the response objects? (same for the `conn.urlopen` test)

tests/modules/test_urllib3.py
galbash: maybe better to test both cause the body may be handled differently

epsagon/events/urllib3.py
```diff
@@ -107,13 +107,20 @@ def update_response(self, response):
         self.resource['metadata']['response_body'] = None
         full_url = self.resource['metadata']['url']
 
-        if not is_payload_collection_blacklisted(full_url):
+        if (
+                response.tell() > 0 and
```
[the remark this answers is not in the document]
henprtz: Nice catch.

tests/test_utils.py
henprtz: Done

tests/modules/test_urllib3.py
henprtz: Done

tests/modules/test_urllib3.py
henprtz: Fixed

tests/modules/test_urllib3.py
henprtz: 1. Fixed
2. Added another test

tests/modules/test_urllib3.py
```diff
@@ -5,55 +5,98 @@
...
+    assert (len(wrapped_data) == 0)
     # But, data will be stored in the `data` object
+    assert (len(wrapped_response.data) > 0)
+    # Compare un-instrumented vs instrumented data
+    assert (len(data) == 0)
     assert (len(response.data) > 0)
-    urllib3_event = trace_transport.last_trace.events[-1]
-    # Payload will be collected and stored inside the event data
-    assert(len(urllib3_event.resource['metadata']['response_body']) > 0)
+
```
adavidai, line 76: `assert (response.data == wrapped_data)`
Should we also check the peek attribute of the responses to make sure they are the same?

tests/modules/test_urllib3.py
```diff
@@ -5,55 +5,98 @@
...
+
+def test_data_capture_with_pool_manager_pc_false(trace_transport):
```
adavidai, line 78: Can be a parameterized test with a preload_content arg:
```
@pytest.mark.parametrize("preload_content", [true, false])
def test_data_capture_with_pool_manager(preload_content):
```
Instead of two different tests, no?
(and compare the wrapped and non wrapped `.read()` as well as `.data`

tests/modules/test_urllib3.py
henprtz: You can't do that since the data is different

tests/modules/test_urllib3.py
henprtz: I'll fix this

tests/modules/test_urllib3.py
adavidai: I'm sure we can find a site for testing that returns the same data each time. 
Such a test would be very powerful, and would make sure the original response is not harmed by the instrumentation

henprtz: Agree. Done

adavidai (approved):

PR #364, opened 2021-07-05
fix(utils.py): show original import error

ranrib:
Instead of getting this:
`Unable to import module 'epsagon': Failed to import module: index`
Get this:
`Unable to import module 'epsagon': Failed to import module: lambda_function, Error: No module named 'mod'`

(this will expose the real import error to the user)

### DIFF
--- a/epsagon/utils.py
+++ b/epsagon/utils.py
@@ -280,8 +280,8 @@ def import_original_module():
 
     try:
         return __import__(module_path), module_path, handler_name
-    except ImportError:
-        raise ImportError('Failed to import module: {}'.format(module_path))
+    except ImportError as ex:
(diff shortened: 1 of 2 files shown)


ophiryael:
The locking of the version might help with the tests but it means that we don't fully support new versions of fastapi and we won't have the same metadata when using newer versions. Maybe it's worth mentioning it in the readme or just fix the issue and ensure we support newer versions.

ranrib:
> The locking of the version might help with the tests but it means that we don't fully support new versions of fastapi and we won't have the same metadata when using newer versions. Maybe it's worth mentioning it in the readme or just fix the issue and ensure we support newer versions.

That's true - I'll open a ticket for it for the team to handle separately

ophiryael (approved):

PR #366, opened 2021-07-20
feat(fastapi): added ignored endpoints

sagivr2020:


adavidai (commented):
Looks good 👍 
How about adding a test for the ignored endpoint functionality to the existing fastapi tests?

sagivr2020:
According to "run_tests.sh" file - pytest ignored FastAPI tests.
Therefore I added fastapi_example.py for testing.

adavidai (approved):
Adding a ticket for enabling the tests back, once we do let's add this test as well

PR #376, opened 2021-08-12
Requests stream response support cs 3

maorlx:


epsagon/events/requests.py
```diff
@@ -65,15 +69,42 @@ def __init__(self, wrapped, instance, args, kwargs, start_time, response,
...
+        if not data:
+            return data
+        try:
+            return json.loads(data)
+        except ValueError:
+            if isinstance(data, bytes):
+                try:
+                    return data.decode('utf-8')
+                except UnicodeDecodeError:
+                    return str(data)
```
Alon-Katz, line 102: I would change this to a single return data at the end, then assign data to its value instead of returning at each step

Alon-Katz (approved):

PR #392, opened 2021-11-17
fix(HTTPTransport): No retries will be made when trying to reach url and getting a timeout

Alon-Katz:
When HTTPTransport was getting a timeout while trying to send out traces, it was retrying to send those traces (because of urlib3 configuration), this resulted in application delays when the url is unreachable.

Now the urlib3 in HTTPTransport is configured to not have any retries on failed attempts to reach the url.

### DIFF
--- a/tests/test_transports.py
+++ b/tests/test_transports.py
@@ -1,0 +1,31 @@
+import time
+from pytest_httpserver import HTTPServer
+import pytest
+import urllib3
+from epsagon.trace_transports import HTTPTransport
+from epsagon.trace import (trace_factory)
+
+
+def test_sanity(httpserver: HTTPServer):
+    collector_url = '/collector'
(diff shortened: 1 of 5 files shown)


tests/test_transports.py
```diff
@@ -0,0 +1,30 @@
+import time
+import pytest
+import urllib3
+from epsagon.trace_transports import HTTPTransport
+from epsagon.trace import (trace_factory)
+
+
+def test_sanity(httpserver):
```
maorlx, line 8: would add test_httptransport_sanity (the same for the timeout test)

maorlx (dismissed):

tests/test_transports.py
Alon-Katz: Done

maorlx (approved):

PR #394, opened 2022-01-02
fix(pymongo): extend support for more operations

sagivr2020:
https://epsagon.atlassian.net/browse/CS-119

### DIFF
--- a/epsagon/events/pymongo.py
+++ b/epsagon/events/pymongo.py
@@ -83,18 +83,22 @@ class PyMongoEvent(BaseEvent):
         :param input_args: input from botocore
         :return: None
         """
-        
+
         if self.resource['operation'] == PyMongoEvent.INSERT_MANY:
-            add_data_if_needed(self.resource['metadata'], 'Items', input_args[0])
+            add_data_if_needed(self.resource['metadata'], 'Items',
+                input_args[0])
 
         elif self.resource['operation'] == PyMongoEvent.INSERT_ONE:
-            add_data_if_needed(self.resource['metadata'], 'Item', input_args[0])
+            add_data_if_needed(self.resource['metadata'], 'Item',
+                input_args[0])
 
         elif self.resource['operation'] in PyMongoEvent.FILTER_OPERATIONS:
-            add_data_if_needed(self.resource['metadata'], 'Filter', input_args[0])
-            
+            add_data_if_needed(self.resource['metadata'], 'Filter',
+                input_args[0])
+
             if self.resource['operation'] == 'update_one':
-                add_data_if_needed(self.resource['metadata'], 'New Values', input_args[1])
+                add_data_if_needed(self.resource['metadata'], 'New Values',
+                    input_args[1])
 
 
     def handle_insert_operations_response(self, response):
--- a/epsagon/events/pymongo.py
+++ b/epsagon/events/pymongo.py
@@ -42,10 +42,10 @@ class PyMongoEvent(BaseEvent):
         self.resource['operation'] = getattr(wrapped, '__name__')
 
         if not args:
-            input = list()
+            input_args = list()
 
         else:
-            input = args[0]
+            input_args = args[0]
 
         self.event_id = 'mongo-{}'.format(str(uuid4()))
         self.resource['name'] = instance.name
@@ -60,7 +60,7 @@ class PyMongoEvent(BaseEvent):
                 'Collection Name': str(instance.collection.name),
             }
 
-        self.handle_request_payload(input)
+        self.handle_request_payload(input_args)
 
         if response is not None:
             self.update_response(response)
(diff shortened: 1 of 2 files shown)


epsagon/events/pymongo.py
```diff
@@ -19,6 +19,8 @@ class PyMongoEvent(BaseEvent):
 
     ORIGIN = 'pymongo'
     RESOURCE_TYPE = 'pymongo'
+    INSERT_OPERATIONS = ['insert_one', 'insert_many']
+    FILTER_OPERATIONS = ['find', 'update_one', 'delete_many']
```
maorlx, line 23: would rename to query_operations

epsagon/events/pymongo.py
```diff
@@ -19,6 +19,8 @@ class PyMongoEvent(BaseEvent):
 
     ORIGIN = 'pymongo'
     RESOURCE_TYPE = 'pymongo'
+    INSERT_OPERATIONS = ['insert_one', 'insert_many']
```
maorlx, line 22: great putting that in a const.
Would change it to be a tuple instead of a list

epsagon/events/pymongo.py
```diff
@@ -70,17 +76,38 @@ def update_response(self, response):
...
-
-        if self.resource['operation'] == 'insert_many':
-            self.resource['metadata']['inserted_ids'] = \
-                [str(x) for x in response.inserted_ids]
-        elif self.resource['operation'] == 'insert_one':
-            self.resource['metadata']['inserted_ids'] = \
-                [str(response.inserted_id)]
+
+        if self.resource['operation'] in PyMongoEvent.INSERT_OPERATIONS:
+            for i in range(len(self.resource['metadata'].get('items', []))):
```
maorlx, line 81: I would move each logic (of insert/query operations) to a separated method to handle the response data

epsagon/events/pymongo.py
```diff
@@ -33,30 +35,34 @@ def __init__(self, wrapped, instance, args, kwargs, start_time, response,
...
+                'DB URL': ':'.join([str(x) for x in address]),
+                'DB Name': str(instance.database.name),
+                'Collection Name': str(instance.collection.name),
+            }
+
+        if self.resource['operation'] in PyMongoEvent.INSERT_OPERATIONS:
+            add_data_if_needed(self.resource['metadata'], 'Items', documents)
+
+        elif self.resource['operation'] in PyMongoEvent.FILTER_OPERATIONS:
+            add_data_if_needed(self.resource['metadata'], 'Search', documents)
```
maorlx, line 65: not sure regarding the "Search" key, consider renaming it (we can talk about it)

epsagon/events/pymongo.py
```diff
@@ -33,30 +35,34 @@ def __init__(self, wrapped, instance, args, kwargs, start_time, response,
...
-            'Collection Name': str(instance.name),
-        }
-        add_data_if_needed(self.resource['metadata'], 'Items', documents)
+                'DB URL': ':'.join([str(x) for x in address]),
+                'DB Name': str(instance.database.name),
+                'Collection Name': str(instance.collection.name),
+            }
+
+        if self.resource['operation'] in PyMongoEvent.INSERT_OPERATIONS:
+            add_data_if_needed(self.resource['metadata'], 'Items', documents)
```
maorlx, line 62: I think there's an issue with the document/documents logic (regarding the insert one/many operations)
Also, the documents param is only relevant for specific operations. So, I think I would add some payload colelction here according to the operation (insert many > so it documents, insert one/update one > document, etc)

maorlx (commented):
great PR ! :) 
See my comments, specifically regarding the request payload collection (document/documents)

epsagon/events/pymongo.py
```diff
@@ -70,17 +74,81 @@ def update_response(self, response):
...
+
+        self.resource['metadata']['Filter'] = str(
+                    self.resource['metadata']['Filter']
+                )
+
+        if self.resource['operation'] == 'find':
+            self.resource['metadata']['Results'] = \
+                        list(response)
+
+        elif self.resource['operation'] in ['update_one']:
```
maorlx, line 143: better check if its equal to update_one (if we will have one more similar operation, we can add the "in" check)

epsagon/events/pymongo.py
```diff
@@ -70,17 +74,81 @@ def update_response(self, response):
...
+            self.resource['metadata']['matched_count'] = \
+                        response.matched_count
+            self.resource['metadata']['modified_count'] = \
+                        response.modified_count
+
+        elif self.resource['operation'] in ['delete_many']:
```
maorlx, line 149: same as update_one comment

epsagon/events/pymongo.py
```diff
@@ -70,17 +74,81 @@ def update_response(self, response):
...
+
+
+    def handle_filter_operations_response(self, response):
+        """
+        Handle response data before add to event.
+        :param response: response from botocore
+        :return: None
+        """
+
+        self.resource['metadata']['Filter'] = str(
```
maorlx, line 135: should put it directly as str in the handle_request_payload function

epsagon/events/pymongo.py
```diff
@@ -33,30 +37,30 @@ def __init__(self, wrapped, instance, args, kwargs, start_time, response,
...
-            'DB Name': str(instance.database.name),
-            'Collection Name': str(instance.name),
-        }
-        add_data_if_needed(self.resource['metadata'], 'Items', documents)
+                'DB URL': ':'.join([str(x) for x in address]),
+                'DB Name': str(instance.database.name),
+                'Collection Name': str(instance.collection.name),
+            }
+
+        self.handle_request_payload(input_args)
```
maorlx, line 63: perfect :)

epsagon/events/pymongo.py
```diff
@@ -33,30 +37,30 @@ def __init__(self, wrapped, instance, args, kwargs, start_time, response,
...
-        documents = args[0]
+        self.resource['operation'] = getattr(wrapped, '__name__')
 
-        self.event_id = 'mongo-{}'.format(str(uuid4()))
-        self.resource['name'] = instance.full_name
+        if not args:
+            input_args = list()
+
+        else:
+            input_args = args[0]
```
maorlx, line 48: I think the [0] logic should be done inside the handle_request_payload (according to the operation , you can know what to expect in args)
So basically, you would just send "args" as is to handle_request_payload. The handle_request_payload obviously should be called only if there're any args

epsagon/events/pymongo.py
```diff
@@ -19,6 +19,10 @@ class PyMongoEvent(BaseEvent):
 
     ORIGIN = 'pymongo'
     RESOURCE_TYPE = 'pymongo'
+    INSERT_ONE = 'insert_one'
+    INSERT_MANY = 'insert_many'
+    INSERT_OPERATIONS = (INSERT_ONE, INSERT_MANY)
+    FILTER_OPERATIONS = ['find', 'update_one', 'delete_many']
```
maorlx, line 25: change to tuple

maorlx (approved):
Great :)

PR #397, opened 2022-01-20
fix(trace): show always exception in case of fail to send trace

sagivr2020:


epsagon/trace.py
```diff
@@ -1144,14 +1144,13 @@ def _send_traces(self):
...
-            if self.debug:
-                print(
+            print(
                     'Failed to send trace: '
                     'updating events resource metadata failed: {}'.format(
                         exception
                     )
                 )
-                traceback.print_exc()
+            traceback.print_exc()
```
maorlx, line 1153: I would print the full traceback only on debug

